PR #2, opened 2017-10-18
Dependency Injection support

### LINKED ISSUE #1: Support for dependency injection
Usage of dependency injection to create instances will help to resolve more complicated dependencies. Moreover, it enables proper usage of IDisposable classes.

@michalkovy has already a working version using BoDi (DI used in SpecFlow). It would be nice to get it into the official tickspec nuget release. If you will agree, we can create pull request with the feature.

mchaloupka:
I want to resolve #1.

I have been thinking about using some preexisting DI container but I have not found any suitable one. If the project will be migrated to .NET Core, then it would make sense to use the build-in one, but for now it seems to me better to have a very simple one as part of the project. So, I have added the ability to resolve more complicated types to the ServiceProvider. It is possible to construct a class which depends on another classes (using constructor injection). It is also possible to do simple type or instance registrations, even as part of other steps. The usage is added as an example.

The ScenarioRun and ScenarioGen were modified, but most of the changes are whitespaces (end of line, they were probably inconsistent).


bartelink:
Hey, thanks for this; Sorry I won't have time to do more than a scan until later in the week. One quick thing, also to assist other reviewers, `ScenarioRun.fs` is not showing diffs cleanly in the GitHub UI - it seems the line endings change in the first commit - perhaps a rebase can cure it? Or perhaps a separate reformatting PR might be in order (if any of this is not sounding like a 2 minute job that you'd prefer to do yourself, say the word and I'll do it)

TickSpec/ScenarioRun.fs
```diff
@@ -1,172 +1,173 @@
...
+        | None,None,Some doc -> [|box doc|]
+        | _,_,_ -> [||]
+    let args = Array.append args tail
+    invoke provider meth args
+
+/// Generate scenario execution function
+let generate events parsers (scenario,lines) =
+    fun () ->
+        /// Type instance provider
+        use provider = new ServiceProvider()
```
bartelink: if you made this an optional arg and/or overload it in a different way, perhaps people could feed in a service provider which is instantiated in terms of a DI container (i.e. allowing you to do a 10 line impl which delegates to Autofac. While this should not be the main path and I'm not entirely averse to having a batteries-included experience, we don't want to build a monster either. Having such an extensibility point may allow you to push the InstanceProvider out into the test project rather than relying on it being a part of TickSpec itself.).

I need to think about this aspect a bit more, but my gut sense is that it should be possible to separate out
- the the DI impl side from TickSpec (i.e. default to the boring ServiceProvider unless a specific one is provided and let that be the way toy feed one in
- the DI wiring from the step impls in the Scenario / Steps impl (have not played with it and your example is too small for me to get the full sense of it)

mchaloupka:
Regarding providing your own DI container - at which point do you expect to inject the container? Do you expect it in the:
- StepDefinitions constructor?
- GenerateScenarios method?
- scenario.Action.Invoke method?
- other way?
I guess that there should be an interface as simple as possible to implement. Even the IInstanceProvider is too complex if you will use different container. Maybe it can be just an interface inheriting from IDispose and having Resolve method. Or we can stick to the IServiceProvider interface and checking actively whether it is IDisposable and if so then call the Dispose() method. What do you prefer?

**Done:**
- Downcasts
- Fixed whitespaces (I thought that it was an issue with preexisting files, but it seems that just something went wrong before)

**TBD:**
- Better examples (real usage instead of technical demonstration)
- Ability to provide own DI implementation

bartelink:
Hi @mchaloupka thanks for the quick followup; sorry I'm a bit snowed atm and won't be able to match your pace for the next month ;)

bartelink:
Thanks for cleaning the line endings; your core implementation-incision looks neat now I can see it ;)

Re when to provide `ServiceProvider`, not sure exactly where it would naturally surface if one was to push it out, but I'm thinking in the GenerateScenarios - I'd be looking to stash and/or gen/configure the container in a shared context, possibly sharing some heavy context/cache info across multiple sets of Test Cases. (Also the further this stays away from normal APIs travelled by civilians the better)

In terms of the API, yes, pretty much exactly as you suggest makes sense:-
- At the surface I picture an `?instanceProvider : IServiceProvider option` hidden in a corner
- Under the covers, it could get defaulted to the old base impl, and your impl could live in an example for NUnit
- The key addition I'd make is (again, as you suggested) to make the way you call Dispose not assume a particular provider impl and instead do an `as` cast to `IDisposable` and `Dispose()` iff it is `Dispose`able - IIRC many containers impl `IDisposable` and `IServiceProvider` OOTB (think Autofac does)

This core change makes a clean initial PR that's easy to agree/merge.

Then, at the point where we have actual usage extracted from a real test suite (or ideally 3) and have an example of applying the scheme using xUnit in an appropriate fashion, one might include a batteries-included `InstanceServiceProvider`. In my experience, there are various happily working TickSpec implementations out there that are doing their jobs silently which the original author of the base integration has moved on from. If we expose too much of this, we might have to make an ugly breaking change (picture NUnit v4 having some DI facility or tweaking things for parallelism in a manner analogous to how xUnit v2 gained parallelism and changed a lot o assumptions around the edges without actually breaking APIs).

(This also matches a realistic window within which I might more easily extract time to work out an xUnit binding!)

.gitignore
```diff
@@ -237,6 +237,7 @@ ModelManifest.xml
 
 # Paket dependency manager
 .paket/paket.exe
+.paket/Paket.Restore.targets
```
bartelink: can you gitignore and add `nuget nunit` via paket (perhaps in a `group build` to make sure it's not transitive in the Nuget) rather than including the .dll in the repo please ? I'd really like to avoid including binaries and/or otherwise making things heavy if possibel.

mchaloupka: I can see that @michalkovy opened a pull request with that. If you accept it, then I will merge it here and add it even for the newly added projects.

bartelink: @mchaloupka there are 3x PRs. You are correct that the Paket and NUnit stuff in that overlaps and covers this issue, but his stuff is reliant on your core work so I'm keen for us to work to get this trimmed down to a minimal impl which meets the goal and allows him to build his stuff from your base. @michalkovy it does seem to make sense to carve off your NUnit / Paket bits right away and then get this ready to merge after that ?

bartelink: Seems `Paket.Restore.targets` should actually be added, not `.gitignore`d, see https://fsprojects.github.io/Paket/faq.html#What-files-should-I-commit

michalkovy: Probably that information isn't up to date, see [this documentation issue in Paket](https://github.com/fsprojects/Paket/issues/2840#issuecomment-336080872)

bartelink: @mchaloupka The NUnit via Paket changes have been merged, so it might be useful to merge or rebase now

@michalkovy can your Functional DI examples work based on this ? I have yet to look at the latest of this branch or your actual branch, but it would be great if we can establish common abstractions. hopefully with as little as possible in the core itself.

michalkovy: I am not sure that I understand the question, but I will try to reply. I fully reused what Milos did, I just missed one method in the container. Milos has there RegisterInstanceAs<'TInterface>method which calls instances.Add(typeof<'TInterface>, instance). That wasn't good enough for me as I need to be able to replace instance in container, too. Hence I added method RegisterOrReplaceInstanceAsType which calls instances.[t] <- instance. However, when I think about more my implementation actually breaks the concept that you would replace DI container how Milos wanted, we will think it through.

I have forward integrated the Functional DI branch. Half of the code is Milos's DI branch.

michalkovy: I am not sure that I understand the question, but I will try to reply. I fully reused what Milos did, I just missed one method in the container. Milos has there RegisterInstanceAs<'TInterface>method which calls instances.Add(typeof<'TInterface>, instance). That wasn't good enough for me as I need to be able to replace instance in container, too. Hence I added method RegisterOrReplaceInstanceAsType which calls instances.[t] <- instance. However, when I think about more my implementation actually breaks the concept that you would replace DI container how Milos wanted, we will think it through.

I have forward integrated the Functional DI branch. Half of the code is Milos's DI branch.

bartelink: @michalkovy Thanks for the response. I did assume you were building on Milos' work.

Regarding merging this, my main concern for any impl is that TickSpec's core should touch `IServiceProvider` and `IDisposable` only.

Then the specific DI container (perhaps a batteries included basic impl can live in core to serve the samples but in principle I should be able to wire in a random version of e.g. Autofac if I want) can should be fed in via the top level of the API.

If we get something together that fits that description, then you guys can both build examples of your approaches which would work against an unmodified TickSpec NuGet.

My overall idea/plan is to get either this PR or a new one to do basic plumbing that's able to cater to both needs and then review the approaches on top of that via separate PRs.

michalkovy: I think Milos achieved that but unfortunately I need for the functional DI to work the TickSpec core to be able to change instances in container by the TickSpec core. Or, what I could do is to have a custom container in TickSpec core that I know within the container - but I don't feel that it would be better for custom DI framework. So, one option would be to not use IServiceProvider but have a slightly bigger interface that would also contain the method to change an instance in the container. What do you think?

bartelink: I have not looked at your code but its definitely best to keep the interface from the core to the container minimal. IServiceProvider and IDisposable have clear known semantics that we dont need to write an article about. So I'd suggest adding a function or single method interface to model the exact semantic of how stuff is emitted would be best (or could you model it as a post step Event that something external can Subscribe to ?). Growing a big state ball with lots of proprietary methods is the thing I'm seeking to avoid - if we can do it as discrete extension points, there's far better chance we end up providing something useful in the medium term and/or for as yet unforeseen needs.

@mchaloupka what do you think - is there a point in getting minimal `IDisposable` support and extending that or do you see the value being something broader and more complete like Michal's overall approach ?

mchaloupka:
So, there was quite a huge discussion. I will try to address all mentioned parts.

When looking at the pull request made by @michalkovy, I really liked the way as it allows a functional way of writing tests. It enables easier way how to share a state/context/whatever without having mutable variables.

On the other hand, I completely agree with @bartelink that it should be possible to provide own way of providing some standard DI container. But I do not see a reason why to keep the old implementation as the new one behaves exactly the same for the previously supported scenarios. It just supports more. So, the new implementation should be the default. However, in the samples (and in our real test cases) we actually do not need the type registration. So, the default implementation can be simplified to have just two methods: "Resolve" and "RegisterInstance". These two methods will be latter used also in the @michalkovy approach. If you will want anything more, then you can use some more advanced DI container.

So, what I suggest is that I will do the next:

- Simplify the IInstanceProvider interface
- According to that simplify also the ServiceProvider
- Provide a way how to optionally add own IInstanceProvider

michalkovy:
I fine with the approach that @mchaloupka proposed

bartelink:
@mchaloupka Yes, that sounds great.

>Simplify the IInstanceProvider interface

I assume by this you mean that the TickSpec engine will

a) use a supplied IServiceProvider
b) Dispose instances at the right time
c) emit via a "SaveInstance" of some kind

Nothing is popping into my head about where to fall wrt to:
a) derive from IServiceProvider adding a `RegisterInstance` (breaking CQS)
b) take a pair of `resolve` and `register` functions and do away with the usage of `IServiceProvider` (which kinda made sense for the same reason many DI containers implement IServiceProvider)
c) have a `ITickSpecContainer` with a bespoke and clear protocol

Do you have ideas or are you intending to experiment and report back?

Examples/ByFeatures/DependencyInjection/Dependency.feature
```diff
@@ -0,0 +1,13 @@
+﻿Feature: Feeding dog
```
bartelink: Am liking this scenario, it strikes a good balance between insane-artificial and showing some of the power.

Examples/ByFeatures/DependencyInjection/DependencySteps.fs
```diff
@@ -0,0 +1,44 @@
...
+        | Large -> 400
+
+type PersonSteps(instanceProvider: IInstanceProvider, bowl: DogBowl) =
+    [<Given>]
+    member this.``I have a (.*) dog`` (size) =
+        match size with
+        | "large" -> instanceProvider.RegisterInstance typeof<Dog> (Dog(Large))
+        | "medium" -> instanceProvider.RegisterInstance typeof<Dog> (Dog(Medium))
+        | "small" -> instanceProvider.RegisterInstance typeof<Dog> (Dog(Small))
+        | _ -> Assert.Fail("Unsupported dog size")
```
bartelink: use `invalidArg` or an assert that includes the value

Examples/ByFeatures/DependencyInjection/DependencySteps.fs
```diff
@@ -0,0 +1,44 @@
...
+
+    [<When>]
+    member this.``I fill the dog bowl with (.*)g of food`` (amount:int) =
+        bowl.FoodAmount <- amount
+
+type DogSteps(bowl: DogBowl, dog: Dog) =
+    [<When>]
+    member this.``The dog eats the food from bowl`` () =
+        bowl.FoodAmount <- bowl.FoodAmount - dog.AmountToEat
```
bartelink: model the domain - bowl.Eat ;) (and guard invariants!) (not really)

Examples/ByFeatures/DependencyInjection/FeatureFixture.fs
```diff
@@ -0,0 +1,23 @@
+﻿module TickSpec.NUnit
+
+open System.Reflection
+open NUnit.Framework
+open TickSpec
+
+let assembly = Assembly.GetExecutingAssembly()
```
bartelink: consider separating this out like in xUnit2 with a `TickScenarios.fs` (that a NuGet package could emit - ideally with a type that you don't have to derive from and can just use as a one line) and a `Features.fs` which in one line lets you expose a set of scenarios to the runner. (Only if you have appetite, this is unimportant and I will prob do it as a backport as part of the work of porting your example to xUnit -- but it is nice to have a new user not have a mangle of tech code with no comments and a base class they need to understand)

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
 open System
 open System.Collections.Generic
 open System.Diagnostics
+open System.Reflection
 
+/// Provides an instance provider for tests
+type IInstanceProvider =
+    inherit IDisposable
```
bartelink: in general people don't do this (and/or it causes a mess when they do) - `IDisposable` is an orthogonal interface and we should not force people to implement it

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
...
 open System.Collections.Generic
 open System.Diagnostics
+open System.Reflection
 
+/// Provides an instance provider for tests
+type IInstanceProvider =
+    inherit IDisposable
+
+    /// Resolves an instance of desired type
+    abstract member Resolve : Type -> obj
```
bartelink: I think IServiceProvider makes sense to use as a base interface for this and leave the name as it was - it has clear documented semantics

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
...
+
+    /// Registers an instance for a type (if there is already a registered instance, it will be replaced)
+    abstract member RegisterInstance : Type -> obj -> unit
```
bartelink: is it a problem for this to be generic (is it much more messy in the codegen) ? Or is it a specific choice from an interface design perspective?

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
...
+        let alreadyRequested =
+            typeStack
+            |> List.tryFind (fun x -> x = t)
+
+        match alreadyRequested with
+        | Some _ -> raise (InvalidOperationException(sprintf "Circular dependency found when resolving type %O" t))
+        | None -> ()
+
+        match t with
+        | t when t = typeof<IInstanceProvider> -> self :> obj
```
bartelink: in line with the [ISP](https://en.wikipedia.org/wiki/Interface_segregation_principle) it would be good to be able to indicate one only consumes and does not provide by asking for IServiceProvider ?

Having said that, this will probably complicate the impl a bit here

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
...
+        member this.Resolve (t: Type) =
             getInstance t
+
+        member this.RegisterInstance (t: Type) (instance: obj) =
+            instances.[t] <- instance
+
+    interface IDisposable with
+        member this.Dispose() =
+            instances.Values
+            |> Seq.choose (fun x ->
```
bartelink: I would merge the iter and the choose

Or 

```
for i in instances.Values do
    match i with
    | :? Disposable as disposable -> disposable.Dispose()
    | _ -> ()  
```

TickSpec/ServiceProvider.fs
bartelink: supporting `fun unit -> 't` and `fun 't -> unit` (maybe that gets messy) would allow one to bind to an F# type:-

```type MyFeature(resolve, register) =
    let x : DogBowl = resolve()
    let doStuff () =
        register <| SmallDog()
```

that might get us out of the interface game?
Or e.g. [delegate factories](https://autofac.readthedocs.io/en/latest/advanced/delegate-factories.html) can allow us to pass args to ctors.
Why am I mentioning this?

I'm concerned that having a very specific interface and semantics in this container will make it much harder to ever support other semantics. Thinking....

TickSpec/ServiceProvider.fs
```diff
@@ -3,21 +3,94 @@
...
+                instances.Add(t, instance)
+                instance
+
+    /// Creates an instance if there was none for a speicified type (and remembering the stack of types being resolved)
+    and createInstance (t:Type) (typeStack: Type list) =
+        let constructors =
+            t.GetConstructors()
+            |> List.ofArray
+
+        let (_, widestConstructors) =
```
bartelink: am thinking, oh noes, we is writin' another DI container (though I agree that some form of batteries being included is important (how does this implf fare under .NET Core, did we say it was OK ?))

TickSpec/ScenarioGen.fs
```diff
@@ -468,8 +471,16 @@ let defineRunMethod
...
     gen.BeginFinallyBlock()
+    // Execute after scenario events
     afterScenarioEvents |> emitEvents
     gen.EndExceptionBlock()
+    gen.Emit(OpCodes.Leave_S, exitOuter)
+    gen.BeginFinallyBlock()
+    // Dispose the ServiceProvider
+    gen.Emit(OpCodes.Ldarg_0)
+    gen.Emit(OpCodes.Ldfld, providerField)
+    gen.Emit(OpCodes.Callvirt, typeof<IDisposable>.GetMethod("Dispose"))
```
bartelink: can this be made conditional please ? i.e. Dispose it iff it impls IDisposable

bartelink:
Do you or @michalkovy have a plan to have step method results that are not `Unit` automatically get registered in the container? i.e. is there a need for the code to be aware of the full IServiceProvider abstraction ? I guess I need to go look at the PR (the reason I'm asking is that I'm considering doing and merging the small bit to expose the serviceProvider arg with a view to giving myself enough to do an xUnit and/or autofac example (I guess I can do that on my own branch...))

bartelink:
`master` now has the whitespaces changes and some extra linting stuff now.

bartelink:
I also spiked the conditional disposal in https://github.com/fsprojects/TickSpec/tree/feature/di, if you want to nab that

mchaloupka:
Yeah, I have yet to do:

- Conditional disposal
- Custom DI provider in the emitted variant
- Address other (I would say minor comments)

Hope I will be able to do it today.

TickSpec/ServiceProvider.fs
mchaloupka: I wanted to make the interface in the exact way how it is used by TickSpec. TickSpec will not use the generic variant. So, I tried to make the interface as small as possible. I can easily add generic variants into it.

TickSpec/ServiceProvider.fs
mchaloupka: So, you prefer to have two separate interfaces for registration and resolving and you would not be able to ask for the complete container, just for one of the two interfaces? I can do that, but I do not understand the reason. Regarding ISP, both methods are for container manipulation, storing and retrieving so from this principle it seems ok to me... Again, I have no problem doing it, although I am not 100% sure, that it is worth it.

Regarding the two methods - they look nice and I can imagine, that we may want to add it as an addition. However, it does not look to be right in terms of DI, because what type you use when you call register in that case? SmallDog, Dog or IDog? I think that it is good to be explicit in this case.

TickSpec/ServiceProvider.fs
mchaloupka: If we move under .NET core then I would replace it by the one used in .NET Core. However, if you look on the implementation now, it is not DI, it is just the previous provider with recursion (so a type may need a different type) and with the ability to prepare an instance from outside. It seems to me ok, to use this approach., to have such simple container here and provide the ability to replace it by something real.

bartelink:
@mchaloupka Thanks. Please don't feel you need to rush it - as long as you're prepared to iterate, we'll get there, and I'll feed in the xUnit samples during it.
I've applied zero thought to how instancing works and am not married to it - I just wanted it to be DRY
Ping me when you have a core that admits a the two input functions and/or however you are modelling that and I'll try to fulfill it via Autofac

TickSpec/ServiceProvider.fs
bartelink: You are correct to model the extensibility point directly as TickSpec wants it and have wrapping elsewhere. (i.e. there's a lot to be said for the IServiceProvider method definition)
My concern is that if its also supposed to be user facing, then there should be a clean idiomatic F# route that allows you to work with type inference in the main

TickSpec/ServiceProvider.fs
bartelink: Sorry, I misformatted the example and omitted the `type` definition.
The container should be substitutable and is not the interesting bit. For the bulk of what TickSpec wants, IServiceProvider is sufficient.
If we build a way for return values to be registered, then we need a way for that to work.
But in general an author of a Feature type should be able to ask for either 
- 80% of the time (ideally via ctor args), values like normal DI, by supplying params as opposed to explicit calls
- 5 % of the time registering stuff to share with other steps
- 15% of the time interacting in fancy ways - people in that space will probably be best off using Autofac or their favorite container.
Baking an interface which does the register and the resolve doesn't really help anyone in the above categories as I see it. Does this make any sense ?

Re whether the signatures are generic or have an explcit type arg, for resolve that gets very ugly very fast. Also if a container ha preregistered stuff, you dont want to do explicit calls
```
type MyFeature(kennel : Kennel, resolve, register) =
    let x : DogBowl = resolve()
    let doStuff () =
        register <| SmallDog()
    let letTheDogsOut() =
      kennel.letThemOut ()
```
If we do implicit register of the return value, then you're in the land of having to make sure the types line up anyway.

None of this precludes also having a way to supply a `Type` too, but in general, tpying `typeof<X>` is not something I'd want in normal DI code, so why would I want to be doing that in step defintiions ?

bartelink: The other thing is regarding splitting registering and resolving, many containers (such as Autofac) make a very deliberate contrast between registering and resolving from a CQS perspective (I dont even know how you do runtime registers in Autofac as in general, it bakes a graph, but assume there is a way)

From an ISP perspective I'd argue that you're only very rarely doing both registering and resolving in any given context

TickSpec/ServiceProvider.fs
bartelink: Do you mean the one in ASP.NET Core ? Is that in a separable package ?

mchaloupka: Yes.

mchaloupka:
So @bartelink , I have it in a form where I am happy with it. It provides the ability to use custom provider which is not needed to be IDisposable. I have addressed all other minor comments.

It seems to me that there is only one point without final decision and that is the form of IInstanceProvider in terms of usage in steps. To be honest, I think that if you want to write it in a functional way, then the way proposed by @michalkovy should be used. It is the cleanest way from functional point of view. If you do not use it "functionally" then it seems to me fine to have some accessible container with the ability to register and retrieve an instance. If you want, I can simply split it into two different interfaces - IServiceProvider and something like IServiceSaver. And you can ask for the one you need.

bartelink:
@mchaloupka Thanks for the work; I won't get to review immediately but am looking forward to it. Hopefully @michalkovy can do the same.

I definitely would like the register and resolve interfaces to be separated, but maybe that's not critical at this point if you guys can agree a common base from which both approaches can function; i.e. if you guys want to move toward merging then let's do that and not wait until I get around to spiking an autofac integration and porting the sample to xUnit.

TickSpec/ServiceProvider.fs
bartelink: Cool - that might make a pretty decent LCD container to target given the degree to which one should ideally be employing containers in an acceptance testing scenario.

TickSpec/ServiceProvider.fs
michalkovy: Also function could be use instead of fun + match

Examples/ByFeatures/DependencyInjection/DependencySteps.fs
```diff
@@ -0,0 +1,49 @@
+﻿module DependencySteps
```
michalkovy: my preference would be to have the file split by step classes - it will than be closer to real use and as I plan to use the same test in functional injection it will be better comparison

michalkovy:
I am fine with this implementation

bartelink:
@mchaloupka seeing whitespace diffs above (its saying 1000+ lines changed); can you have a peep and let me know when it's good to review?

mchaloupka:
@bartelink Argh, I do not understand when it happens. Anyway, feel free to review the other parts. I guess that it currently happened only in the single file. I will fix it tomorrow. The change in the generation is just the usage of lambda func to create the provider and then the conditional Dispose call (I took it from your branch).

bartelink:
👍 Will have a look later with a local diff tool

Examples/ByFeatures/DependencyInjection/Dependency.feature
```diff
@@ -0,0 +1,13 @@
...
+
+    Scenario: Feeding dog with enough food
```
michalkovy: this should be Scenario Outline instead of Scenario

michalkovy:
@mchaloupka I have realized in the dog feeding example that it may be nice to register DogSize instead of Dog and so show a little bit more how DI works and can help. See my changeset https://github.com/michalkovy/tickspec/commit/e7596f79d83959aa509888b2550101a5259fbe5a in FI variant. What do you think?

bartelink:
V nice @michalkovy this is all starting to take shape - need to rename the PersonSteps at some point ;)

mchaloupka:
So, the whitespaces are fixed.

Have I addressed all the comments? Or is there anything else you want me to do before merging?

bartelink:
Thanks Milos - unfortunately I'm majorly snowed atm and thus have not gotten around to looking yet :( will do as soon as I can

bartelink:
Hey @mchaloupka have not forgotten this, but insanely busy this week so still have not made time to look, sorry :(

bartelink:
Looking at this now... First step is rebasing on master, which is slightly messy due to a combo of lots of commits here and me having added a commit regarding IDisposable. I'm reluctant to just shave that off though... Setting up a gitter room https://gitter.im/fsprojects/TickSpec

mchaloupka:
Yes, the history is somehow messy but the diff seems to be ok. Moreover, I expect squash merge so it does not matter.

mchaloupka:
Closing the pull request as it was merged as part of other pr.

PR #4, opened 2017-10-29
NUnit via Paket

michalkovy:
Removal of current NUnit binaries and replacement by use of Paket.

The base for changes is VS2017 pull request (so it contains now both changes).


bartelink:
See #2, #3 I'd definitely for reffing an in-date version of NUnit via Paket so this is an ideal quick win PR. Can you elaborate on the options as to what platforms to target implied by that? Ideally we don't go requiring more things without adding facilities to the system (i.e., I'm keen to see core still only require F# 3.1 and .NET 4.0 unless that's really unavoidable as it can paint us into corners. While this might seem like a lot of work for no reward, for me the example is set by [unquote](https://github.com/SwensenSoftware/unquote) - you simply don't run into contexts where it's the problem as a lot of attention has clearly been paid to minimizing dependences).

So, in short, is there a way to update NUnit without requiring 4.5 ?

bartelink:
Seems `Paket.Restore.targets` should actually be added, not `.gitignore`d, see https://fsprojects.github.io/Paket/faq.html#What-files-should-I-commit

michalkovy:
This pull request should be now ready for merge.

bartelink:
Thanks. Will look+merge later.

PR #5, opened 2017-10-29
Functional Dependency Injection

michalkovy:
Today TickSpec slightly fails in the functional approach as it is hard to avoid mutable fields and you can't make easily steps as a module with tick functions (without class) when you want allow parallel tests. Also as you use more field read/write approach instead of more pure function approach in steps the code overall is less functional. 

I think I have fixed that by allowing a new style of writing steps:

1. Step functions can have a return value. When a value V of type T is returned then the V is registered to container under type T. If next time other function returns value X of the same type T then V is replaced by X
2. Functions can have more arguments. Any extra argument is resolved using dependency container

This new style of coding should not affect the other styles. The only cons I can think of is that if you want to write by the old style and you make a mistake (extra argument or unintentional return of a value) then it may be slightly harder to catch it.

This pull request contains these other pull requests:
- Dependency Injection
- VS2017
- NUnit via Paket


bartelink:
As discussed in the other PRs, this holds lots of promise but
- we need to do one thing at a time
- we def need to get stuff compiling in 2017, move to paket and in-date NUnit
- we need to not add extra deps / complexity to core that's avoidable
- we need to make sure we can generate NuGet packages along the way
- for this DI approach, I went to look a the diffs (I have had similar thoughts in the past), but
    - there are some changes I hope can be avoided (and stuff that's under discussion with with mchaloupka in #2  - we need 'more real' examples) which make it hard to see the forest for the trees
    - for me, it's all about proving >1 person has made this work in practice - there's little point in just merging it until we get to that point so it may be valuable for others to work off this branch and converge on an approach
   - based on these two points I've yet to dig in in depth; please don't consider this a no - I have not even looked in any meaningful way.

As noted, I'm a bit busy to shepherd this this month unless we work in very small steps.

michalkovy:
I have updated the pull request to contain up-to-date @mchaloupka Dependency Injection changes. I have created a copy of his feeding dog example to functional injection examples and converted it to the functional injection style - it gives a nice comparison. I have also added the standard Stock example in FI style to compare. More, I also updated readme to contain that style as extra option as I think it is better default for F# coding? What do you think @bartelink ?

README.md
```diff
@@ -46,6 +47,24 @@ let [<Then>] ``I should have (.*) black jumpers in stock`` (n:int) =
     Debug.Assert(passed)
 ```
 
+# Step definitions (F# without mutable field)
```
bartelink: This looks fantastic 👏

bartelink:
The example in the readme looks great. Will look at code in due course (diffs are not showing clearly atm)

michalkovy:
I have merged Milos's last changes again and I have fixed the line feed, so diff should be good now

bartelink:
Have not had time to look today. Am I correct in inferring that you are happy to finalize your stuff after we finalize and merge Milos's ?

If you feel your work on this in any way helps with deciding 50/50 calls in the other PR, the more you feed in there the better (but the fact you are basing off his is likely to have already validated that to a large degree)

michalkovy:
Sure, let's first merge Milos's PR first, I am already fine with what he did, he adjusted his version so that I don't have any change to the ServiceProvider.fs - I just reused his DI for different purpose.

mchaloupka:
Closing as this branch is already merged as part of other pr.

PR #6, opened 2017-10-30
Removal of Silverlight support

michalkovy:
Silverlight didn't win the battle of technologies and it isn't even included in Visual Studio 2017. It may be best to remove support from TickSpec to make the repository smaller to manage and to be able to build TickSpec from command line when Visual Studio 2017 is on a machine only.


README.md
```diff
@@ -11,7 +11,7 @@ Easily execute the behaviour against matching F# tick methods, attributed C# or
 # Installation
 
 Simply reference TickSpec via [Nuget](https://www.nuget.org/packages/TickSpec/), download the assembly or build the project from source.
-TickSpec works in Visual Studio 2008, 2010 & 2012 and supports both .Net and Silverlight projects.
+TickSpec works in Visual Studio 2008, 2010 & 2012.
```
bartelink: Perhaps add something like `Historically, Silverlight was supported; this support and the related examples were removed in 2017 (but remain in the commit history for the archeologically inclined) ?

TickSpec/ScenarioGen.fs
```diff
@@ -348,13 +348,7 @@ let defineStepMethod
     // Emit arguments
     let ps = mi.GetParameters()
     let zipped = Seq.zip args ps
-#if SILVERLIGHT
-    // For instance methods get instance value from service provider
```
bartelink: IME it can be helpful to have these variations about when porting to more constrained platforms, i.e. .NET Standard. Arguably the merging / removal of this could should wait until that point (only being devil's advocate really though)

bartelink:
OK, #7 is merged - I'm in two minds about removing the `#if SILVERLIGHT` (I know .NET Standard 2 support is a different ballgame and deleting code is a good thing, but is something in a `#ifdef` really code :troll: ?); perhaps #4 is a better one to do next ? I'll let you decide...

michalkovy:
Thanks, I have added the comment to README regarding Silverlight.

I have made a research regarding .NET Core by compiling it for .NET Core, I though it may be best to test the hypothesis. So, I was able to compile TickSpec for .NET Core 2 with a small changes and I was able to run the a test scenario. The output of my research:
1. we can safely remove the defines for Silverlight, there doesn't seem to be any problem
2. the emit works with one exception, I didn't find quickly a replacement for ModuleBuilder.DefineDocument and ILGenerator.MarkSequencePoint which both are likely used for the linking with lines for debugging feature file - which is the main reason for emit. Nevertheless, I didn't invest much.

bartelink:
That all sounds great - @ptrelford you don't happen to have played with about with .NET Core in this space?

@michalkovy Any reason not to merge as-is; do you envisage turning this into a .NET Core PR :trollface: (even if you were I think completing this now makes sense)

michalkovy:
Sure, I am for merging it as-is.

Sorry, I won't change it to .net core PR 😊

bartelink:
Merged, thanks!

PR #7, opened 2017-10-30
Remove of VS2010 and VS2008 projects/solutions

michalkovy:
I don't think anybody will use Visual Studio 2008 or 2010 today to build TickSpec.

I think it is better not to have these project in as when we do changes to TickSpen we don't test if it works in these old versions, so it is better not to have them instead of having them potentially broken.


README.md
```diff
@@ -11,7 +11,7 @@ Easily execute the behaviour against matching F# tick methods, attributed C# or
 # Installation
 
 Simply reference TickSpec via [Nuget](https://www.nuget.org/packages/TickSpec/), download the assembly or build the project from source.
-TickSpec works in Visual Studio 2008, 2010 & 2012 and supports both .Net and Silverlight projects.
+TickSpec works in Visual Studio 2017.
```
bartelink: Can you make it say 2015 too  please ?- I will verify. I think that's still an important target fora variety of users.

bartelink:
Thanks; this looks good and I can't imagine anyone arguing so won't canvas broader agreement.

I just tried it with 2015 - it compiles and runs the Xunit tests so I say :shipit:.

Can you tweak the README (as an `amend` - I prefer not to squash-merge reasonably curated PRs) to mention support for 2015 and do a final boarding check please ? 

I'll merge when you give the go-ahead.

michalkovy:
Done, I have added VS2015 to the readme.

bartelink:
Thank you for having the courage to call this out and do it!

PR #11, opened 2017-11-15
Configure TickSpec build for AppVeyor CI

mchaloupka:


LICENSE
```diff
@@ -0,0 +1,176 @@
+                                 Apache License
```
bartelink: @ptrelford what should the license be ?

build.bat
```diff
@@ -1,20 +1,35 @@
...
-cd ..\..
\ No newline at end of file
+msbuild TickSpec.sln /p:Configuration=Release /m
+
+if errorlevel 1 (
+  echo "Build failed"
+  exit /b %errorlevel%
+)
+
+if %1.==. (
```
bartelink: can you quote the comparands please ?

build.bat
```diff
@@ -1,20 +1,35 @@
...
+    set version="1.0.1.1"
+) else (
+    set version="%1"
+)
+
+.paket\paket.exe pack packed_nugets --version %version% --symbols
+
+pushd %~dp0packed_nugets
+%NuGetTool% pack ..\Nuget\NUnit\TickSpec.NUnit.nuspec -Version %version%
+%NuGetTool% pack ..\Nuget\xUnit\TickSpec.xUnit.nuspec -Version %version%
```
bartelink: Maybe we should drop this - few people use xUnit v1 and it's not helpful to send people the wrong way (we might need to unlist the package ... or maybe we should log an issue to do a package that has all the helpers ready to go for xUnit v2 without the sample - obviously this was all done before Paket so for example there are source files in the package which are intended to be edited, which is not a good pattern)

LICENSE
mchaloupka: I took it from here: http://tickspec.codeplex.com/license

bartelink: Hm, how did I drop it, apologies!

bartelink:
Feel free to merge when you're ready (I prefer not to squash  if the individual commits are clean)

build.bat
mchaloupka: Done.

build.bat
mchaloupka: I commented it out. We should probably consider how to provide such thing (does it really need a nuget)? For example, the NUnit variant will not work even for NUnit3. I agree that handling it as a separate issue is good way to go.

LICENSE
mchaloupka: I have just noticed it when converting the nuspec to paket.template. I also fixed the link to the project there.

build.bat
bartelink: Yeah - if you look on the DI branch, I have a cleaner way to to the NUnit linking (based onthe xUnit style) which does not bind to the specific version of NUnit (i.e. no base type).
Will focus on getting DI merged first but happy to discuss the approach on an Issue if you want

PR #12, opened 2017-11-23
Fix nuget dependency

mchaloupka:
This pull request targets the configuration of the nuget created.

The problem was that paket takes the nuget configuration according to "how we want to build the project". However, we want to build project with older FSharp.Core. That does not mean that it will not work with a newer one. So, we should not prevent usage in projects with newer FSharp.Core.


bartelink:
lgtm; merge when you see fit

PR #15, opened 2018-03-15
enhance build + publish fake target

cboudereau:
Hello @sergey-tihon and @bartelink, I have done 2 commits (one for migrating to fake + paket and the last one for the publish target).

I have rewritten the build process in order to be able to reproduce the build locally without appveyor by using fake + paket.

I have also 
- added a RELEASE_NOTES.md to publish with the nuget (this will display the release notes on nuget.org)
- Turning paket to magic mode

Let me know if there is something wrong


cboudereau:
The appveyor build is failing due to the parameter. Could you remove the version because know it has been moved to RELEASE_NOTES.md ?

Thanks!

.gitignore
```diff
@@ -236,12 +236,15 @@ GeneratedArtifacts/
 ModelManifest.xml
 
 # Paket dependency manager
-.paket/paket.exe
 .paket/Paket.Restore.targets
 paket-files/
+packages/
```
mchaloupka: This is already ignored, right?

build.fsx
```diff
@@ -0,0 +1,95 @@
...
+            BuildNumber |> Option.defaultValue "0" 
+            |> sprintf "%s.%s" ReleaseNotes.TickSpec.AssemblyVersion
+        [ Attribute.Version fileVersion
+          Attribute.FileVersion fileVersion ]
+        |> AssemblyInfoFile.UpdateAttributes asmInfo)
+
+Target "Build" <| fun _ -> build Build.setParams Sln
+
+Target "Test" <| fun _ ->
+    let examples x = Build.rootDir </> "Examples" </> x
```
mchaloupka: Is it possible to run the tests without listing all the assemblies here? It will be easy to miss it when adding new tests.

build.fsx
```diff
@@ -0,0 +1,95 @@
...
+    [ examples "ByFeature/DependencyInjection/bin/Release/DependencyInjection.dll"
+      examples "ByFeature/FunctionalInjection/bin/Release/FunctionalInjection.dll"
+      examples "ByFramework/NUnit/CSharp.NUnit/bin/Release/NUnit.cSharp.dll"
+      examples "ByFramework/NUnit/FSharp.NUnit/bin/Release/NUnit.FSharp.dll"
+      examples "ByStyle/Attributes/bin/Release/Attributes.dll"
+      examples "ByStyle/Functional/bin/Release/Functional.dll" ]
+    |> NUnit3 (fun p ->
```
mchaloupka: NUnit2 is used.

bartelink:
Thanks for doing all this @cboudereau (and @mchaloupka for the reviewing)!

I think its reasonable to switch to NUnit3 for the runner (assuming that like xUnit 2 it has backcompat for running tests authored against v2) - the supporting NUnit2 intent is more wrt whether the lib itself can be used to author an NUnit2 test and/or whether a user of the lib can just do a paket update without getting dragged into a black hole.

I won't be able to review in full until later as in the middle of stuff, but if a merge to `master` is needed to validate anything I'm good with that (squashing is not necessary but I'd prefer not too many 'fixing things' commits to get onto `master` itself).

cboudereau:
Hello, @bartelink, @mchaloupka here is the last version including : 
.gitignore has been fixed
test assemblies are discovered for nunit and xunit.

As @bartelink says, NUnit runner 3 can be used for test assemblies referencing Nunit2.

Could you merge the pull request + update appveyor by removing the version parameter (the version is located into the RELEASE_NOTES.md file) causing build problems in order to see if it is ok ?

.gitignore
cboudereau: fixed

build.fsx
cboudereau: Yes fixed into the last commit

build.fsx
cboudereau: Nunit3 runner supports both

mchaloupka:
> Could you merge the pull request + update AppVeyor by removing the version parameter 
The version parameter is part of yaml file. Change it in your pull request to get a green build. It is not configured anywhere in AppVeyor.

cboudereau:
@mchaloupka, yes you are right. Now it is ok.

The AssemblyInfo.fs is updated automatically by the script before compiling the nupkg.

If you want to publish the nuget you can now run build.bat "publish" and setting the nugetapi key as environment variable (its name is displayed into the console error message).

mchaloupka:
Looks good from my perspective. Just one minor thing: before your changes the nuget produced from build had version with the build number (so it was increasing with every build). Now, it has only three static numbers, so the builds produce the nugets with same version. It would be probably nicer to have all four numbers so the produced nuget packages will be unique (anyone can take it from the build and use it).

cboudereau:
@mchaloupka, the nuget number is exactly the semver supplied into the RELEASE_NOTES.md and the buildnumber appear in assembly version and fileversion. You can check it by extracting the nupec file from the appveyor artifact. So in the publish target, it take only the .symbols.nuget that contains nuspec with release notes version and notes with assemblies patched according to version from release notes (ie 1.1.0) plus build number (ie : 1.1.0.50).

Is it ok for you ?

mchaloupka:
Okay. I was thinking about the scenario if someone will take the nuget from build so he will see that a newer version was released and that it is the different than the last released (without unpacking the nuget). However, there may be additional improvements for nuget releasing (like automatic releasing from build according to some trigger), so we can do them separately.

It looks good, thanks for your work. @bartelink : Do you have any comments? If not, let's merge this pull request.

bartelink:
I have not actually looked at the code but know about RELEASE_NOTES and the general discussion and you both are happy I'm confident I will be too.

I won't get to look in detail until the weekend as I need to focus on something else for the remainder of today (sorry for not getting to look/comment yesterday evening - I had a glance and it all seems fine).

I'm happy for it to be merged as-is or not as you guys see fit without my explicit review ;)

sergey-tihon:
Thank you @cboudereau.
Added my 50cent with build for Mono and Travis config

cboudereau:
Thank you @sergey-tihon I was a little short to do this so now all CI are ok! That's a great thing!

TickSpec.sln
```diff
@@ -46,7 +46,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "._", "._", "{51B6BE98-A176-
 	ProjectSection(SolutionItems) = preProject
 		appveyor.yml = appveyor.yml
```
bartelink: travis too?

appveyor.yml
```diff
@@ -1,7 +1,7 @@
 image: Visual Studio 2017
 build_script:
-  - cmd: build.bat %APPVEYOR_BUILD_VERSION%
-test: on
+  - cmd: build.bat
+test: off
```
bartelink: can we 'just' make them  work instead ?

build.bat
```diff
@@ -1,33 +1,15 @@
 @echo off
 
-.paket\paket.bootstrapper.exe
```
bartelink: see L3- bootstrapper needs to run before paket.exe (and build.fsx refs some of the stuff that installs)

build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+if test "$OS" = "Windows_NT"
+then
+  # use .Net
+  .paket/paket.exe restore
```
bartelink: needs to bootstrap too

paket.dependencies
```diff
@@ -12,4 +12,12 @@ nuget xunit.runner.visualstudio ~> 2 version_in_path: false
 nuget Unquote
 
 # Example container to illustrate integration
-nuget Autofac
\ No newline at end of file
+nuget Autofac
+
+group build
+framework: auto-detect
```
bartelink: indent please (by 4 spaces like in rest of repo)

build.fsx
```diff
@@ -0,0 +1,120 @@
...
+    Paket.Pack (fun p ->
+        { p with
+            BuildPlatform = "AnyCPU"
+            MinimumFromLockFile = true
+            Symbols = true
+            OutputPath = Build.nuget 
+            ReleaseNotes = String.concat System.Environment.NewLine ReleaseNotes.TickSpec.Notes
+            Version = ReleaseNotes.TickSpec.NugetVersion } )
+
+Target "Publish" <| fun _ ->
```
bartelink: can we add a comment line or two please (or should it live in a diff file ?)

build.fsx
```diff
@@ -0,0 +1,120 @@
...
+    |> XUnit2.xUnit2 (fun p -> 
+        { p with 
+            ToolPath = Build.rootDir </> "packages/build/xunit.runner.console/tools/net452/xunit.console.exe"
+            ForceAppVeyor = AppVeyor.BuildNumber |> Option.isSome })
+
+Target "Nuget" <| fun _ -> 
+    Paket.Pack (fun p ->
+        { p with
+            BuildPlatform = "AnyCPU"
+            MinimumFromLockFile = true
```
bartelink: Can the paket publish template get deleted ? have we figured out whether the default meta for the package remains as it was before (and e.g. has not grown?)

paket.lock
```diff
@@ -21,3 +21,35 @@ NUGET
...
+      NUnit.Extension.NUnitV2ResultWriter (>= 3.5)
+      NUnit.Extension.TeamCityEventListener (>= 1.0.3)
+      NUnit.Extension.VSProjectLoader (>= 3.7)
+    NUnit.ConsoleRunner (3.8)
+    NUnit.Extension.NUnitProjectLoader (3.6)
+    NUnit.Extension.NUnitV2Driver (3.7)
+    NUnit.Extension.NUnitV2ResultWriter (3.6)
+    NUnit.Extension.TeamCityEventListener (1.0.3)
+    NUnit.Extension.VSProjectLoader (3.7)
+    Paket.Core (5.151.4)
```
bartelink: this all looks like a lot of stuff, and paket bootstrap + restore is responsible for getting fake so I think this should not live in here (or in the .dependencies, ideally?) I maye def be wrong; if this is for a good reason, please disregard

appveyor.yml
cboudereau: Tests are now ran from rhe build script

build.bat
cboudereau: Paket.bootstrap has been renamed to paket.exe (magic mode). So there is no need to call the bootstrap first.

build.fsx
cboudereau: The dependencies are not really ised since it was hard code into the tickspec.paket.template (I think you could separate example from rhe lib by using another group)

paket.lock
cboudereau: Sure but now the runner + unit test framework are getting from the build script not by the appveyor runtime. For nunit I follow the reco to use the nunit runner console but it came with this dependencies.

In general it is better to have nuget tools rhat come from rhe build so we can share the script for appveyor + travis

appveyor.yml
bartelink: duh, makes sense!

build.bat
bartelink: ah, ok, missed that, great

paket.lock
bartelink: OK, this is all rooted in my misunderstanding of the paket bootstrapping. As long as the batch files are managing the load of the fake+runners+dependences (i.e. paket.core for packing) before the running of the fake (which I now see it does consistently), this all looks good.

build.fsx
bartelink: My only concern is to avoid duplication and/or confusion between the `paket.template` and this block here, i.e. the ideal is if this block is minimal in size (though one wouldnt want to publish any other way than this, the emphasis should be on removing as much as possible from the template). This is not a blocker would be nice to get done for once and for all

RELEASE_NOTES.md
```diff
@@ -0,0 +1,3 @@
+#### New in 1.1.0 (Released 2018/03/15)
+* enhance build process with fake+paket
```
bartelink: can it mention the "Add dependency management support from @mchaloupka and @michalkovy" please?

build.fsx
cboudereau: If you run the target, the env vars is displayed into the error message.

paket.dependencies
cboudereau: done

build.fsx
cboudereau: done

RELEASE_NOTES.md
cboudereau: done + mine

cboudereau:
I have refactored the build script + reduce nunit internal dep  and updated the release notes by including your contribs too.

cboudereau:
Hello @mchaloupka could you merge the pull request please ?

PR #16, opened 2018-09-30
Migration to .NET Standard 2.0 + .NET 4.5

### LINKED ISSUE #8: Support for dotnet core 2.0
Do you have any plans to add support for dotnet core 2.0?
Notw it fails because It can't load the features by reflection.

michalkovy:
I have migrated TickSpec to .NET Standard 2.0. I have made all projects to output both .NET Standard 2.0 and .NET 4.5 binaries. 

This resolves #8 

The solution doesn't support debugging feature files when .NET Standard assembly is used, it is still supported with .NET Framework assembly.


mchaloupka:
Please fix the builds. Feel free to consider migration to FAKE 5 if it makes it easier for you.

TickSpec/TickSpec.fsproj
```diff
@@ -1,63 +1,19 @@
...
-    <RootNamespace>TickSpec</RootNamespace>
-    <AssemblyName>TickSpec</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
-    <Name>TickSpec</Name>
-    <RunPostBuildEvent>OnBuildSuccess</RunPostBuildEvent>
-    <TargetFrameworkProfile />
-    <TargetFSharpCoreVersion>4.3.0.0</TargetFSharpCoreVersion>
+    <TargetFrameworks>netstandard2.0;net45</TargetFrameworks>
+    <Title>TickSpec</Title>
+    <Description>
```
mchaloupka: Is this used even when paket is used to create the nuget packages?

michalkovy: Paket pack doesn't seem to deal well with multi-target projects (the resulting nuget contains one target only). It seems that dotnet pack is the advised solution, so I migrated to that. These elements are used.

Examples/ByFeature/CustomContainer/paket.references
```diff
@@ -1,4 +1,6 @@
 ﻿Autofac
 FSharp.Core
 unquote
-Xunit
\ No newline at end of file
+Xunit
+xunit.runner.visualstudio
+Microsoft.NET.Test.Sdk
```
mchaloupka: Why this is needed?

Examples/ByFeature/DependencyInjection/paket.references
```diff
@@ -1,2 +1,4 @@
 ﻿FSharp.Core
-NUnit
\ No newline at end of file
+NUnit
+Microsoft.NET.Test.Sdk
+NUnit3TestAdapter
```
mchaloupka: Am I right that you are migrating NUnit2 to NUnit3? I am fine with that but it may make sense to have both in examples (if it does not mean significant additional work). What do you think?

build.proj
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mchaloupka: Why is this file needed?

paket.dependencies
```diff
@@ -1,25 +1,41 @@
 redirects: on
-framework: auto-detect
+framework: net45, net452, netstandard2.0, netcoreapp2.1
```
mchaloupka: It is some kind of defect that it needs to be listed?

Do we want to reduce the count (like having only net45 or net452 and the .NET Core ones)?

Examples/ByFeature/CustomContainer/paket.references
michalkovy: It is package that is added byt default to .NET Core test projects, see e.g. https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-nunit

True is that I don't know full details, I know that I had some issues without it.

There was an issue https://github.com/Microsoft/vstest/issues/986 which contains partially reason why it is needed but also try to avoid need for it for some cases.

Examples/ByFeature/DependencyInjection/paket.references
michalkovy: Yes, I am migrating to NUnit 3. The main reason is that NUnit 2 neither supports .NET Standard nor it is maintained library - basically it is just older version of the library which is over 4 years old by this time. I could potentially compile it for .NET Framework only and do some special treating in fake build script but I don't see it worth. I wouldn't recommend anybody today to start using TickSpec with NUnit 2 and if somebody has an old NUnit 2 based project it will still work for him the same way.

It makes to me much better sense to make mapping for other current frameworks, e.g. MSTest v2.

build.proj
michalkovy: I have removed it to simplify it, it isn't necessary. However, the file is the official way how to make "dotnet build" to build using fake, see:  https://github.com/FakeBuild/fake-bootstrap/tree/paket_clitool

bartelink:
Just wow; congrats.

Minor side points:
- The release notes need an update and a version bump (For me it warrants 2.0).
- The commits seem reasonably purposeful - would not argue with a non-squashed merge.
- I'm seeing a failing test in `CustomContainer` on my local 15.9.0 Preview 2 when I run all tests OOTB in Debug `Message: System.InvalidOperationException : Cannot decide which constructor to use. The type has multiple constructors with the same maximum number of parameters: System.IntPtr`

build.proj
mchaloupka: Yeah, but we have also other bootstraps there (the batch and shell script). I don't see a reason to have a bootstrap commited if we do not expect it to be used. Or, do we use it?

Examples/ByFeature/DependencyInjection/paket.references
mchaloupka: Sounds good to me.

Examples/ByFeature/CustomContainer/paket.references
mchaloupka: Sounds good, thanks.

TickSpec/TickSpec.fsproj
mchaloupka: Ok, fine for me. The created package looks good.

build.proj
michalkovy: The other bootstraps were using it. However, as I wrote, I got rid of it. The only benefit was that .NET Core developers could be used to using dotnet build command. But they should be able to run a batch file instead.

mchaloupka:
I can confirm that I am able to see locally the same issue as @bartelink regarding the failing test. It seems to me like that the AutoFac wiring is not working properly (just a guess).

Moreover, when I press run all tests, VS claims that 70 tests passed, 1 failed and 1 not runned (I don't understand why, it is in the CSharp.NUnit and it was runned). The tests from xUnit.FSharp projects are missing completely.

michalkovy:
Yes, I know about the issue with CustomContainer test doesn't work and you can see that I am intentionally skipping the test in fake build. 

It is really broken in the path where IL generation isn't used, so .NET tests succeeds and .NET Core fails. However, it has nothing to do with .NET Core, it was broken before when you use the path of code where IL generation isn't used.

When I was looking at the code it felt to me that problem is that the default container is already taken when custom container is replaced - so it uses default container instead of custom one. I expected that we will fix this in separate pull request as it may be harder, feel free to fix it.

Regarding the missing test, VS behaves weirdly, when you open solution and build for first time it will correctly show just 71 tests but when you run the tests it will add one additional test which should not be there and which isn't executed, it is the CSharp.NUnit test which is called Test. It can be something in NUnitWiring.cs which I rewrote simplified way for NUnit 3 but I don't see the issue, any idea?

build.fsx
```diff
@@ -79,50 +72,55 @@ module Build =
...
-NUnit.assemblies >> NUnit.run 
->> XUnit.assemblies >> XUnit.run
-|> Target "Test"
-
-Target "Nuget" <| fun _ -> 
-    Paket.Pack (fun p ->
-        { p with
-            BuildPlatform = "AnyCPU"
-            MinimumFromLockFile = true
-            Symbols = true
```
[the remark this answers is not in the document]
michalkovy: Unfortunately there is. The reason is that I ma seeing this defect when trying to pack with symbols:
https://github.com/fsprojects/Paket/issues/3272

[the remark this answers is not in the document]
mchaloupka: Ideally create an issue that we will potentially return to it oneday when it will be resolved.

mchaloupka:
* Have you seen also the missing xUnit.FSharp tests?
* I have no idea why the extra "not runned" test appears in test explorer. It somehow looks like an issue in the Test Explorer but maybe we are triggering it somehow as it does not happen in other cases. I can try to research it tomorrow.
* Regarding the test - let's create an issue and solve it later. You are right that this pull request is big enough. However, do you have any suggestions how to ensure that both methods will work? Like running the tests in both variants? Or, are you already doing that?

mchaloupka:
Ah, you removed the project which used the xUnit1 - I am fine with accepting the same argument as for nUnit2. The xUnit2 project is the same, it just uses the latest xUnit. However, in that case please rename xUnit2 project to just xUnit because the number does not have any meaning.

The failed test is fixed in branch bugfix/DelayReadingOfServiceProviderValue. I will start pull request as soon as you merge these changes, so the test can be fixed.

The issue with the not-started test seems like an issue in Test Explorer. Seems like VS tries to guess which test is there without compiling and then it does not remove it when running. Still, in my opinion it is not anyhow a blocker to merge this. Even if it would be an issue, we can handle it later.

michalkovy:
I have done the changes:
- made nuget packaging to again take version and release notes from release notes file (requested by @mchaloupka)
- renamed xUnit2 project to xUnit (requested by @mchaloupka)
- updated release notes to version 2.0.0 and put there this upgrade as the change (requested by @bartelink )

mchaloupka:
All comments were addressed, so I merged the changes. If we find any issues then we will address them separately.

PR #17, opened 2018-10-05
Updated file parsing for feature files

mchaloupka:
I have update the feature file parsing.

Main motivation was:
* Support meaningful combination of example tables
* Support tags on example tables
* Support tags for the whole feature

To achieve that, I have in fact rewritten the parsing almost completely. It follows the same concept but I added some types so there is a definition what goes from line parser to block parser and what then goes to feature parser. This should simplify future possible enhancements / fixes.

I have also added a unit test for parsing. I plan to add at least one more (to cover more possibilities). However, to start the discussion now, let's discuss without the additional test.


michalkovy:
Is there any benefit from removing supporting feature files e.g. like this one?:
```
Story: Returns go to stock

Scenario 1: Refunded items should be returned to stock
Given that a customer previously bought a black sweater from me
And I have three black sweaters in stock.
When they return the black sweater for a refund
Then I should have four black sweaters in stock.

Scenario 2: Replaced items should be returned to stock
Given that a customer previously bought a blue garment from me
And I have two blue garments in stock
And three black garments in stock.
When they return the blue garment for a replacement in black
Then I should have three blue garments in stock
And two black garments in stock.
```
(indexed scenarios and story instead of feature)

If somebody uses TickSpec today or would like to use it with examples on blogs they will need to figure out that we stopped supporting it.

mchaloupka:
You are probably right. I can improve the parsing to support this again. When doing the changes I followed the gherkin spec. However, you are right that the samples on the web exists in this format so it makes perfect sense to return it there. What I don't like in that case is that the name the gherkin keywords will be part of the scenario name. Other thing I do not like is that if you write feature description in a way that "Scenario 1 is about this and scenario 2 about this" than it will parse the description as scenario start. I will change it as you are right that the documentation (and blogs) shows it that way.

One more thing I changed but it is probably not visible anywhere. Before my changes it was possible to have shared examples with one tag, written like: 
```
Shared examples of @all:
```
After my changes you have to write
```
@all
Shared examples:
```

Advantage of my approach is that you can write more tags and even on more lines. Moreover it works the same way also for examples which are part of scenario.

mchaloupka:
I have:
* Returned back the changes so feature files with numbered scenarios are supported again
* Added extra unit test to verify the behavior of items under steps (table, bullet list, doc string)
* Improved exceptions related to parsing

I think that right now it is in final form. I will wait for closing of the pull request #16 - after that the forward integration will remove lots of changed project files (as I was just enabling local builds by them). Moreover, I expect that even the builds will be working after this forward integration.

TickSpec/LineParser.fs
```diff
@@ -47,16 +47,14 @@ let startsWith (pattern:string) (s:string) =
...
-            "^\s*Scenario Outline:\s(.*)"
-            "^\s*Scenario Template:\s(.*)"
-        ]
-        |> Seq.map (tryRegex s)
-        |> Seq.map (Option.map (fun t -> ScenarioLine (t.Trim())))
+        [ "Scenario"; "Story" ]
+        |> Seq.map (fun x ->
+            if trimmed |> startsWith x then Some trimmed
+            else None)
+        |> Seq.map (Option.map (fun t -> ScenarioLine t))
```
bartelink: `Seq.map >> Seq.choose id` -> `Seq.choose`
`Seq.map >> Seq.map (Option.map f)` -> `Seq.choose >> Seq.map f`

```
|> Seq.choose (fun x ->
    if trimmed |> startsWith x then Some trimmed
    else None)
|> Seq.map ScenarioLine
```

Examples/ByFeature/FunctionalInjection/Stock.feature
```diff
@@ -1,6 +1,6 @@
 ﻿Feature: Refunded or replaced items should be returned to stock
 
-Scenario: Refunded items should be returned to stock
+Scenario 1: Refunded items should be returned to stock
```
bartelink: Not sure this makes sense in terms of the intention of the Gherkin design. Could just suggest using `Scenario: 1. ...` and not encourage people to get into rippling auto-numbering edits

TickSpec/LineParser.fs
mchaloupka: Sadly, it is not possible to use `ScenarioLine` like this. However, I have modified it so it looks better. Thanks for your comment.

Examples/ByFeature/FunctionalInjection/Stock.feature
mchaloupka: I would prefer to separate it from this pull request. @michalkovy is correct that all blog posts shows this particular usage. We can decide that we do not want to support it anymore, but it can be easily driven as separate pull request (the change is small)

paket.lock
```diff
@@ -1,156 +1,156 @@
 REDIRECTS: ON
-RESTRICTION: || (== net45) (== net452) (== netcoreapp2.1) (== netstandard2.0)
+RESTRICTION: || (== net20) (== net40) (== net45) (== net452) (== netcoreapp2.1) (== netstandard2.0)
```
michalkovy: looks like you had or have some .net 2 and .net 4 projects at the time of lock creation - try to stay at net45 or 452 where necessary

mchaloupka: It was a missed Mono project. I have removed it as it should not be used anymore.

mchaloupka:
I have forward integrated the pull request. There are many small changes, so I would go with squashed merge if you do not have any additional comments.

See my responses on your comments. They should be addressed.

Any more comments @michalkovy or @bartelink ?

bartelink:
What kind of a merge are you intending to do ? I'd be open to doing a rebase and merge, but the best thing to do if that's the plane is to rebase now (to remove the merge commits) (and consider squashing some minor ones into the previous as a 'fix'). I'll try to have a read through later as I've not yet looked at the readme changes, but my general stance from sniffing the commit messages as they flew by and looking at the quality of the work and reviewing all along is that there's no reason to hold off on merging if you guys are happy.

TickSpec/LineParser.fs
```diff
@@ -39,49 +44,54 @@ let tryRegex input pattern =
...
+    |> Option.map (fun t -> FeatureLine (t.Trim()))
+let (|ScenarioLine|_|) (s:string) =
+    let trimmed = s.Trim()
+    let matches =
+        [ "Scenario"; "Story" ]
+        |> Seq.choose (fun x ->
+            if trimmed |> startsWith x then Some trimmed
+            else None)
+    if matches |> Seq.isEmpty then None
+    else ScenarioLine (matches |> Seq.head) |> Some
```
michalkovy: the above seems complicated, original or the following feels simpler
```FSharp
[ "Scenario"; "Story" ]
|> Seq.tryPick (fun x -> 
    if trimmed |> startsWith x
    then ScenarioLine trimmed |> Some 
    else None) 
```

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
+        // Cross-join tables with different columns
+        |> combinations
+        |> Seq.map processRow
+        |> Seq.choose id
+        |> Seq.groupBy (fun (_,r) -> r)
+        |> Seq.map (fun (row, taggedRows) ->
+            taggedRows
+            |> Seq.fold (fun tags (t,r) ->
+                Seq.append tags t
+            ) Seq.empty
```
michalkovy: Consider: According to https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-function-parameter-application I think it is weird to continue argument on the same scope, I would prefer e.g.:
```FSharp
|> Seq.fold
    (fun tags (t,r) -> Seq.append tags t)
    Seq.empty 
```
you have above in new code much bigger Seq.fold uses with big lambda expression and a similar closing + adding last argument. The formatting guidelines tell "However, if the body of a lambda expression is more than one line, consider factoring it out into a separate function rather than have a multi-line construct applied as a single argument to a function."

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
+                    match newStateMap with
+                    | None -> None
+                    | Some s ->
+                        let newStateTags = stateTags @ rowTags
+                        Some (newStateTags, s)
+            ) (Some (List.Empty, Map.empty))
+
+        examples
+        |> Seq.map (fun exampleBlock ->
+            exampleBlock.Table.Header |> List.sort,
```
michalkovy: I find the commas on end of lines in the middle of big block somewhat non-intuitive (you have it on more places). https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#formatting-tuples requires tuples to be parenthesized. Then to me having contents of the second part huge makes tuple hart to recognize

michalkovy:
I have put in a few minor formatting comments but I am fine with it in general.

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
+                        rowMap
+                        |> Map.fold (fun state key value ->
+                            match state with
+                            | None -> None
+                            | Some map ->
+                                let existingValue = map |> Map.tryFind key
+                                match existingValue with
+                                | None -> Some (map.Add (key, value))
+                                | Some x when x = value -> Some map
+                                | _ -> None) (Some stateMap)
```
bartelink: when doing a fold, as per the F# coding guidelines, it can work well to do

`(Some stateMap,rowMap) ||> Map.fold (fun ...` so the correlated bits sit together and the initial state doesnt ahve to hand at the end

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
-                scenarioName,tags,steps,Some(Array.append exampleTables sharedExamples)
+        // Union tables with the same columns
+        |> Seq.groupBy (fun (h,_) -> h)
+        |> Seq.map (fun (header,tables) ->
+            header, tables |> Seq.collect (fun (_, rows) -> rows) |> Seq.toList)
+        |> Seq.toList
+        // Cross-join tables with different columns
+        |> combinations
+        |> Seq.map processRow
+        |> Seq.choose id
```
bartelink: `map >> choose id` -> `choose`

TickSpec/FeatureParser.fs
bartelink: Again, the Seq.empty at the end is the problem for me (and I don't like the fun on next line

suggest 
```
(Seq.empty, taggedRows) ||> Seq.fold (fun (row, taggedRows) ->
    Seq.append tags t)
```

(If the line ended between 80-120 I'd consider keeping it on one line too)

That guide also mentions an important reason for the pipe-frward and/or double pipeforward is for the type inference to flow correctly

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
+        let processedStep =
+            match step.Step with
+            | GivenStep s -> replace combination s |> GivenStep
+            | WhenStep s -> replace combination s |> WhenStep
+            | ThenStep s  -> replace combination s |> ThenStep
+
+        let bullets, table, doc =
+            match step.Item with
+            | Some (BulletsItem b) -> Some (b |> List.toArray), None, None
+            | Some (TableItem t) -> None, Some (new Table(t.Header |> List.toArray, t.Rows |> Seq.map List.toArray |> Seq.toArray)), None
```
bartelink: break this out  - in general if you find this many parens, its a signal to split into temps

TickSpec/FeatureParser.fs
```diff
@@ -1,92 +1,131 @@
...
+                else
+                    fun _ -> scenario.Name
+
+            exampleCombinations
+            |> List.mapi (fun i (tags, combination) ->
+                let name = nameFunc i
                 let steps =
-                    Seq.append background steps
-                    |> Seq.map toStep
+                    background @ scenario.Steps
```
bartelink: I think this will do a List append, Seq.append will have better perf so best to leave

TickSpec/LineParser.fs
```diff
@@ -39,49 +44,54 @@ let tryRegex input pattern =
...
+let (|BackgroundLine|_|) s =
     tryRegex s "^\s*Background(.*)"
-    |> Option.map (fun t -> IsBackground)
+    |> Option.map (fun t -> BackgroundLine)
 let (|GivenLine|_|) s =
-    tryRegex s "^\s*Given\s+(.*)"
-    |> Option.map (fun t -> GivenLine t)
+    tryRegex s "^\s*Given\s(.*)"
+    |> Option.map (fun t -> GivenLine (t.Trim()))
```
bartelink: (above)
```
let (|Trim|) (s:string) = s.Trim()
```
Then
```
|> Option.map (function Trim t -> GivenLine t)
```

TickSpec/LineParser.fs
```diff
@@ -39,49 +44,54 @@ let tryRegex input pattern =
...
-let (|Row|_|) (s:string) =
+    tryRegex s "^\s*But\s(.*)"
+    |> Option.map (fun t -> ButLine (t.Trim()))
+let (|TableRowLine|_|) (s:string) =
     if s.Trim().StartsWith("|") then
         let options = System.StringSplitOptions.RemoveEmptyEntries
         let cols = s.Trim().Split([|'|'|],options)
-        let cols = cols |> Array.map (fun s -> s.Trim())
-        Row cols |> Some
+        let cols = cols |> Array.toList |> List.map (fun s -> s.Trim())
```
bartelink: `Seq.ofArray >> Seq.map >> List.ofSeq`
or 
`[for s in cols -> s.Trim()]`
or (using actie pattern above)
`[for (Trim s) in cols -> s]`

TickSpec/LineParser.fs
mchaloupka: Thanks.

TickSpec/FeatureParser.fs
mchaloupka: Thanks, I have changed it a little. Does it look better?

TickSpec/FeatureParser.fs
mchaloupka: Improved. Do you see any other occurrences you would like to improve?

Examples/ByFeature/TaggedExamples/FeatureFixture.fs
```diff
@@ -0,0 +1,39 @@
+﻿module TickSpec.NUnit
+
+open System.Reflection
+open NUnit.Framework
+open TickSpec
+
+/// Inherit from FeatureFixture to define a feature fixture
```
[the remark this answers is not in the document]
mchaloupka: Removed, thanks.

Examples/ByFeature/TaggedExamples/FeatureFixture.fs
```diff
@@ -0,0 +1,39 @@
...
+[<TestFixture>]
+type FeatureFixture () =
+    [<Test>]
+    [<TestCaseSource("Scenarios")>]
+    member this.TestScenario (scenario:Scenario) =
+        if scenario.Tags |> Seq.exists ((=) "ignore") then
```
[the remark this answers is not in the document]
mchaloupka: Changed, thanks.

TickSpec.Tests/TagsAndExamples.feature
```diff
@@ -0,0 +1,26 @@
+﻿@http
+Feature: HTTP server
+
+Background:
+Given User connects to <server>
+
+@basics @index
+Scenario Outline: Tags and Examples Sc.
```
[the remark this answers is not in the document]
mchaloupka: It should have been abbreviation for scenario. Removed. It does not make sense to have it there.

TickSpec/AssemblyInfo.fs
```diff
@@ -2,6 +2,7 @@ namespace TickSpec
 
 open System.Reflection
 open System.Runtime.InteropServices
+open System.Runtime.CompilerServices
```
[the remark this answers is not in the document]
mchaloupka: Ok :)

TickSpec/FeatureParser.fs
mchaloupka: Changed, thanks.

TickSpec/FeatureParser.fs
mchaloupka: You are right.

Examples/ByFeature/TaggedExamples/WebTesting.feature
```diff
@@ -0,0 +1,27 @@
...
+
+    Scenario Outline: Successful login
+    Given I use browser <browser>
+    When I try to login using username <username> and password <password>
+    Then I am logged in
+
+    Examples:
+    | browser       |
+    | Firefox       |
+    | Explorer      |
```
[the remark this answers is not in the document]
mchaloupka: Ok :)

TickSpec/AssemblyInfo.fs
```diff
@@ -10,6 +11,7 @@ open System.Runtime.InteropServices
 [<assembly: AssemblyTrademark("")>]
 
 [<assembly: ComVisible(false)>]
+[<assembly: InternalsVisibleTo("TickSpec.Tests")>]
```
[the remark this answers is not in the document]
mchaloupka: ? :)

TickSpec/FeatureParser.fs
mchaloupka: Thanks, changed.

TickSpec/FeatureParser.fs
mchaloupka: Thanks, changed.

mchaloupka:
I have addressed your comments. Regarding some questions:
* I would use squashed merge. I do not see any particular value in the individual commits as some of them were playing with builds etc. It seems to me that the change itself will look good as a single commit (changes in the parsing + added tests). Do you agree? Or do you want to have several commits (but less than it is now)?
* I think that it is perfect time to merge it now, as we introduce v2 of tickspec. At the same time, with this commit I stayed in the safe zone - there are minimum of breaking changes. Especially in the parts that are somehow documented. In the end, the only breaking change is that the tag on shared examples is written differently (more consistently). I am not aware that this was documented anywhere and it is TickSpec specific functionality anyway. The rest is just added functionality.

TickSpec/FeatureParser.fs
michalkovy: I see that you improved it on other places but I don't see update on this one. Also the guidelines wants the tuples to be parenthesized with only specific exceptions which you didn't change.

Examples/ByFeature/TaggedExamples/WebTesting.feature
```diff
@@ -0,0 +1,27 @@
...
+
+    Scenario Outline: Successful login
+    Given I use browser <browser>
+    When I try to login using username <username> and password <password>
+    Then I am logged in
+
+    Examples:
+    | browser       |
+    | Firefox       |
+    | Edge      |
```
michalkovy: put | at the same positions like lines above and below

michalkovy:
I agree with @mchaloupka that it is perfect time to merge this into v2. I don't see the changes as breaking current functionality with an exception of special rare case that this improves and moves it closer to Cucumber.

bartelink:
All sounds good - in general I personally spend time curating commits; the main reason being to aid my self-reviewing. My main no-no is merges into branches unless it's a really significant merge of two real conflicting work streams (that and single commits with merge commits). Whether it's worth the effort is hard to say - but I definitely have cases where I personally take 5 commits, merge them to 3 on the branch (perhaps reordering them) ... and then squash merge (more frequently I'd do a merge commit) ! Most people would consider this insane ;)

In general, I'd say your code change seems reasonable, but then I've never tried to grok the parsing logic before your changes. One thing about squash merges is that you should consider to minimize the diffs by doing a [preparatory refactoring commit or PR](https://martinfowler.com/articles/preparatory-refactoring-example.html) which is drop dead boring and hence safe, and takes all moving and restyling out of the picture.

If you are, as you say more pinning and codifying things vs inventing significant new syntax, I'm good with merging now, and am not really asking for you to do extra legwork to please me given I have no specific concerns with anything you've done and don't have time to sit down to analyze deeper ;)

Examples/ByFeature/TaggedExamples/WebTesting.feature
mchaloupka: Thanks, done.

TickSpec/FeatureParser.fs
mchaloupka: Resolved, thanks.

TickSpec/FeatureParser.fs
```diff
@@ -49,19 +49,21 @@ let parseFeature (lines:string[]) =
...
+                        Seq.zip exampleBlock.Table.Header row
+                        |> Seq.fold (fun map (header, value) ->
+                                match Map.tryFind header map with
+                                | Some x ->
+                                    let m = sprintf "A single header was specified multiple times in an example block starting at row %d" exampleBlock.LineNumber
+                                    ParseException(m, Some exampleBlock.LineNumber) |> raise
+                                | None -> map |> Map.add header value)
+                            Map.empty
+                    rowMap, exampleBlock.Tags)
+            (headers, rows))
```
bartelink: move L64 to L57 using `||>`, remove parens from 66

mchaloupka: Changed, thanks.

PR #18, opened 2018-10-16
Delay reading of service provider value

mchaloupka:
This pull request fixes the failing test. The problem is with the custom container wiring.

The wiring creates the test suite and then before execution assigns the custom container. Originally, it depended on the way of execution when the container will be retrieved. The generated test using emits had a parameter to get the custom container and therefore it was retrieved when the test was executed. However, when the emits were not used, the value was directly written in lambda - therefore the value was captured when the test lambda was created. I modified it in a way that instead of mutable a reference cell is used. Therefore even in the lambda function the value is captured when the test is executed.


michalkovy:
looks as good fix to me

mchaloupka:
Thanks. @bartelink any comments I need to address before merging?

PR #20, opened 2018-10-17
Example with MSTest V2 usage

michalkovy:
This is a simple example how today MSTest V2 can be used easily with TickSpec. Unfortunately, today MSTest generates dynamic tests under the same GUID (within one test in test explorer).

A RFC is on a pull request to allow choosing how dynamic data are treated, I have put comment about it into the wiring code and I have also commented about our use case in the RFC pull request.

Other option is e.g. to create type provider and generate test methods on compilation time. However, I think it is big benefit of TickSpec over e.g. SpecFlow that it allows generating tests on runtime instead of compilation time. So, I wouldn't invest time in it for now.


Examples/ByFramework/MSTest/MSTest.FSharp/Stock.feature
```diff
@@ -0,0 +1,13 @@
+﻿Feature: Refunded or replaced items should be returned to stock
+
+Scenario: Refunded items should be returned to stock
+	Given a customer buys a black jumper
+	And I have 3 black jumpers left in stock 
+	When he returns the jumper for a refund 
+	Then I should have 4 black jumpers in stock
+
+Scenario: Refunded items should be returned to stock
```
mchaloupka: You have two same scenarios. That is probably unwanted (and it differs from other test frameworks).

mchaloupka: From the same reason it is actually failing to execute the tests using .NET Framework - the emitted type is named according to the scenario name. If you have two scenarios with the same name, then it fails to emit two types with the same name. Is this a thing we want to solve (like randomizing the type name)?

michalkovy: @mchaloupka I fixed the scenario but I have also found that you actually broke it with the parsing changes. The code worked before and the reason is that the code named second instance of such scenario "Scenario: Refunded items should be returned to stock~2". I think we should make it backward compatible. Will you look on it, please?

mchaloupka:
Please proceed with the forward integration (ideally rebase your changes). All builds should be green.

michalkovy:
@mchaloupka I have squashed the commits

PR #21, opened 2018-10-18
Dispose instances in a reversed order of their creation

davehl7:
Sequence of BDD steps can create instances potentially used for dependency injection in next steps. Newer instances can depend on the older ones, so the order of disposing instances should not be random (as in case of Dictionary<_,_>), but strictly deterministic, here in a order reversed to instance creation.

### DIFF
--- a/TickSpec/ServiceProvider.fs
+++ b/TickSpec/ServiceProvider.fs
@@ -1,9 +1,49 @@
 ﻿namespace TickSpec
 
 open System
-open System.Collections.Generic
 open System.Diagnostics
 open System.Reflection
+
+/// Instances storage
+[<Sealed>]
+type InstanceStore () =
+    let mutable instances : list<Type*obj> = []
+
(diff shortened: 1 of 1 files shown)


TickSpec/ServiceProvider.fs
```diff
@@ -1,10 +1,50 @@
...
+                    (key, value) :: (List.rev revHead) @ rest
+                else
+                    inList
+            | head :: rest ->
+                distinctAddRec key value (head :: revHead) rest
+        distinctAddRec key value [] inList
+
+    /// Stores element (skips storing existing, old element removed and new prepended)
+    member this.Store key value =
+        instances <- instances |> distinctAdd key value
```
bartelink: should only store if `:? IDisposable`

mchaloupka: The store serves also as the lookup. So, it has to store everything.

PR #22, opened 2018-10-19
Enable linux build

mchaloupka:
To enable linux build I had:
* Removed usage of paket (because it requires mono)
* Added condition for build that on linux only .net core targets are built
* Update build.fsx to reflect the above

Moreover, I have enabled linux build for AppVeyor. It does not make sense to have multiple linux builds. It is just for consideration how it looks.

Disadvantage of this change is losing paket functionality (synchronization of nuget reference versions). Advantage is simplification of usage as now it is not anything extra needed to build it (except .NET Core SDK on all platforms and .NET 4.5.2 target on Windows). Moreover, it gives option mentioned by @bartelink to have various dependencies for targets.

What do you think about this change?


michalkovy:
It is good to have Linux build working.

I wouldn't stop supporting Mono, I think it is still very valuable platform and usable in scenarios where .NET Core isn't. The proposed change removes it.

I like Paket, I agree that it's value is lower with .NET Core. I like that you fix by this the Linux build. However, it doesn't mean I wouldn't like to reintroduce it once Mono is fixed or there is .NET Core version of Paket. But it is just personal opinion, I am fine with it. And Paket supports different versions of dependencies for various targets (see https://github.com/fsprojects/Paket/pull/3046/files).

mchaloupka:
First of all, this pull request is to open discussion.

To be honest, I think that if we will be able to rework the emits in a way that it will work similarly even under netstandard, then we can remove the net45 target completely. That means that mono will not be needed for anything.

Right now, it does not remove Mono support. It just removes the ability to compile net45 targets under Mono. The nuget still can be used under Mono.

The only question is whether we want to use paket or not. If so, then we need to have working paket under Mono again. Or we need to get paket running without Mono.

michalkovy:
I think it is great that you opened the discussion.

You are right that Mono is still supported for client usage even with this change. However, I also value ability of being compiling and testing TickSpec itself using Mono. Even, if we had .NET Standard version of TickSpec only it is good to support using Mono for both .NET Standard compilation and .NET45 tests running.

bartelink:
You have more knowledge and context, but based on my knowledge, that makes sense so feel free to go ahead with that if you see fit.

build.fsx
```diff
@@ -25,46 +35,50 @@ module Xml =
...
+    module NUnit =
+        let run () =
+            !! "**/*.fsproj"
+            ++ "**/*.csproj"
+            |> Seq.choose (Analysis.projectReferencing "NUnit")
+            |> Seq.iter (fun p -> p |> DotNet.test (fun o ->
+                { o with 
+                    Configuration = DotNet.BuildConfiguration.Release
+                    NoBuild = true
+                    Framework = if Environment.isWindows then None else Some "netcoreapp2.1"}))
```
michalkovy: is this condition still needed after your changes?

build.fsx
```diff
@@ -25,46 +35,50 @@ module Xml =
...
+    module XUnit = 
+        let run () =
+            if Environment.isWindows then
```
michalkovy: is this condition still needed after your changes?

build.fsx
mchaloupka: No, I have removed it.

build.fsx
mchaloupka: It seems that the condition can be similar to the one that was for NUnit. It seems that the netcoreapp2.1 works. However, the net452 does not work. There seems to be some issue in the xUnit library. It looks like it uses .NET remoting and it fails on Mono.

See output of the following build: https://travis-ci.org/fsprojects/TickSpec/jobs/446042499
The issue is already reported:
https://github.com/xunit/xunit/issues/1357

And the author response is as follows:
> I'm almost 100% sure that Mono is not officially supported for dotnet test (we certainly don't make any effort to support it).
Therefore, I will keep the exception for net452 based tests for xUnit on Mono. Moreover, if we encounter Mono issues with nUnit tests then I will return the exception even to the tests for nUnit on Mono. Or we will then consider to remove Mono completely if it will not be supported by the test frameworks anyway.

PR #24, opened 2018-10-22
Feature/nuget symbols

mchaloupka:
Modifying the nuget creation to include sources and generating a symbols package. Moreover, turning on the generation of xml codedoc.


mchaloupka:
Hmm, after trying, I would say that this pull request is actually another reason to remove paket. 

It is not possible to include symbols when paket is used, because of: https://github.com/fsprojects/Paket/issues/3272
It is not possible to have development dependency, as `dotnet pack` does not have that information from paket and paket does not work with multiple targets:
https://github.com/fsprojects/Paket/issues/3191
https://github.com/fsprojects/Paket/issues/2370

So, in my opinion the correct approach is to merge the paket removal and then merge this. Do you agree @michalkovy and @bartelink ?

bartelink:
Go for it!

michalkovy:
I am fine with Paket removal, we can potentially put it back once all the issues are fixed

PR #28, opened 2018-10-31
Enable .NET Framework features again.

mchaloupka:
The .NET Framework features (ScenarioGen and FeatureGen) were commented out for .NET Core but the used definition was missing for .NET Framework. So, the code was actually never used. This is returning it back.

### DIFF
--- a/TickSpec/FeatureGen.fs
+++ b/TickSpec/FeatureGen.fs
@@ -1,6 +1,6 @@
 ﻿namespace TickSpec
 
-#if NET45
+#if !NETSTANDARD2_0
 open System
 open System.Collections.Generic
 open System.Diagnostics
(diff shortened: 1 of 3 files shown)


michalkovy:
Wasn't this actually broken by upgrade to .NET 4.5.2? I think we should either use define NET452 or start using new define which is version independent like NETFRAMEWORK

mchaloupka:
> Wasn't this actually broken by upgrade to .NET 4.5.2?

No

> I think we should either use define NET452 or start using new define which is version independent like NETFRAMEWORK

NETFRAMEWORK sounds good to me.

michalkovy:
I still don't agree with your evaluation what it broke, see: https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks

mchaloupka:
Hmm, I have tried it and maybe I missed one of the definitions. See the updated variant.

PR #29, opened 2018-11-29
Add documentation for step methods returning tuples

mbuhot:
Hi, thanks for creating TickSpec. It's a great library!

I had to look through the source code before I found that step methods returning tuples would allow me to pass multiple values between step methods.

This PR adds some info to the README to mention return tuples.


bartelink:
In general I'd venture a record is a much cleaner way to manage this type of contract. i.e. Explicit is better than implicit; your code does not end up littered with `fst` and `snd` and adding a third element to a tuple is not going to land you with some very confusing type errors.

IMO the readme should probably mention it can be a record and recommend that - 

🤔 Perhaps one of the examples should probably be made to illustrate this if it does not already (perhaps it can illustrate that tuples are types too and work, but recommend records as an alterative in the comment).

bartelink:
Sorry, had completely forgotten about that feature / aspect of tuples - so never mind me!

Wrt docs - a wiki might indeed be nice and/or you might want to blog or do something simple that works like gists yourself.

However I'd say putting even relatively verbose examples toward the bottom of the readme is ultimately going to be the most useful in terms of conveying the capabilities of a system or project to someone browsing at a glance - example: https://github.com/App-vNext/Polly

The bottom line is that any documentation and examples contribution would be fantastic - whatever works for you as a good way to get a baseline in place and let others take it and run is good, so please don't view my subjective preferences as a constraint on how you do it.

PR #33, opened 2018-12-20
Upload nuget package from build

mchaloupka:


RELEASE_NOTES.md
```diff
@@ -1,8 +1,10 @@
-#### New in 2.0.0
+#### 2.0.0-rc1
 * Migration to .NET Standard 2.0 + .NET 4.5 from @michalkovy
 * Improved feature file parser (to support tagged examples) from @mchaloupka
+* Sample usage using Expect framework from @mbuhot
```
bartelink:
```suggestion
* Sample usage using Expecto framework from @mbuhot
```

bartelink: maybe replace "from" with "by" ?

TickSpec/TickSpec.fsproj
```diff
@@ -5,15 +5,16 @@
...
+    <Description>TickSpec is a lightweight Behaviour Driven Development (BDD) framework for .Net.
+Describe behaviour in plain text using the Gherkin business language, i.e. Given, When, Then. Easily execute the behaviour against matching F# 'ticked' methods (let ``tick method`` () = true) or attributed C# or F# methods.</Description>
     <Authors>Phillip Trelford</Authors>
+    <Copyright>Copyright © 2010-9 Phillip Trelford</Copyright>
     <PackageProjectUrl>https://github.com/fsprojects/TickSpec</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/fsprojects/TickSpec/blob/master/LICENSE.md</PackageLicenseUrl>
     <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
     <PackageTags>BDD C# F# .Net</PackageTags>
+    <RepositoryUrl>https://github.com/fsprojects/TickSpec</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
```
bartelink: maybe sourcelink too ? (not necessarily part of this work but this reminded me of that)

mchaloupka: I just found this in the fsproj editor in VS. The source link should be enabled. I have not tested it.

mchaloupka:
It should work in a way (test locally just by setting environment variables to simulate AppVeyor build) that if we tag a commit where the tag will match the nuget version in release notes then it will push the nuget. The nuget API key secret is stored in an environment variable and according to AppVeyor it should be set if and only if the build is triggered for an internal build (that means from branch from this repository, not in a pull request coming from a different repository). So, after merging (I will go with squash) this, I will create tag 2.0.0-prerelease which will be published as nuget.

PR #39, opened 2020-04-23
Install Fake as a CLI tool using a tool manifest

mchaloupka:
I replaced the custom bat/sh scripts with the supported way how to bootstrap fake. After that, I just fixed errors and warnings that appeared in the build (it was quite a long time when it compiled the last time).


TickSpec/TickSpec.fsproj
```diff
@@ -10,16 +10,21 @@ Describe behaviour in plain text using the Gherkin business language, i.e. Given
     <Authors>Phillip Trelford</Authors>
     <Copyright>Copyright © 2010-9 Phillip Trelford</Copyright>
     <PackageProjectUrl>https://github.com/fsprojects/TickSpec</PackageProjectUrl>
-    <PackageLicenseUrl>https://github.com/fsprojects/TickSpec/blob/master/LICENSE.md</PackageLicenseUrl>
+    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
```
bartelink, line 13: should switch to the new form (leave the file, determine the lincense!)
https://github.com/jet/dotnet-templates/blob/master/Directory.Build.props#L8

TickSpec/TickSpec.fsproj
```diff
@@ -38,13 +43,11 @@ Describe behaviour in plain text using the Gherkin business language, i.e. Given
     <Compile Include="FeatureParser.fs" />
     <Compile Include="Feature.fs" />
     <Compile Include="TickSpec.fs" />
+    <None Include="..\LICENSE.txt" Pack="true" PackagePath="LICENSE.txt"/>
```
bartelink, line 46: better to put in a solution folder, no?

bartelink:
:shipit:

TickSpec/TickSpec.fsproj
mchaloupka: That is another option. The url is deprecated. You should either provide a file (as in this pr) or provide name of the license. The file sounds like a better practice to me, as we have it there anyway. Open to suggestions what we want to do regarding licensing.

TickSpec/TickSpec.fsproj
mchaloupka: I prefer to have it in root of the repository, so it is immediately visible. Moreover, I believe that it is expected by the tools.

bartelink: Sorry, I mean I would not add a License to a `.fsproj`, but include it in a `.sln`
Spotted the `PackagePath` now -nmm!

TickSpec/TickSpec.fsproj
bartelink: I'm happy with anything that doesn't produce warnings - I know random legal folk have complained to me about the one above, but it sounds like you've got it covered, so never mind me!

mchaloupka: I am just not touching it. What is the kind of feedback you are receiving. As we are not the original owners, from legal perspective it is not correct to just change it without approval.

bartelink: TL;DR what you have is fine, I am not demanding a change or anything of the sort...

Sorry, we're in violent agreement; not asking for anything to be materially changed at all.
The license is independent of the way the metadata is represented in the package.
I believe `PackageLicense` is a new (well, circa 2016 kinda new) nupkg element which allows certain tooling to more directly infer the package type based on the metadata manifest.
In terms of feedback I'm receiving, my comment was intended to convey: "the way its encoded in the thing I linked to has had tonnes of legal eyeballs on it; it's not like it was randomly thrown together"

PR #44, opened 2021-01-20
Buildfix

mchaloupka:
The build is failing as `FSharp.Core` 5 does not support `net452`. I have forced the project to use older version of `FSharp.Core` to allow it to build. However, I guess that at some point we want to drop the .NET Framework support. We still have some functionality that is specific only to .NET Framework so I have decided not to migrate it immediately.


Examples/ByFeature/FunctionalInjection/FunctionalInjection.fsproj
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\..\netfx.props" />
+  <Import Project="..\..\..\fsharp_core.props" />
```
bartelink, line 4: Should/can't this not go into a Directory.build.props somewhere ?

fsharp_core.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bartelink, line 1: If this is named Directory.build.props then it'll get auto-included in any build of sln files under that folder (the netfx.props should probably also be merged into that?)

bartelink (approved):
Looks good. Up to you whether to apply the suggestion or not

fsharp_core.props
mchaloupka: Makes sense, I have added a condition in the file to apply it only on F# projects.

mchaloupka: Actually, the mono assembly search path has to be referenced as it does not seem to work from the `Directory.Build.props` file. However, I have moved the new bits to it.

bartelink: Ah, sorry for making extra work for you with silly ideas! cool you got it all resolved in the end 👍

mchaloupka: No worries at all. It would be nice if it worked.

PR #47, opened 2021-05-28
feat(PlaceholdersInTables): Placeholder replacements (from Examples) …

deyanp:
…also in tables


deyanp:
Note: Implemented only for tables ..

mchaloupka:
Hi,

thanks for the PR, it looks good! I am fine with merging it as it is but would you mind adding the support also to other step arguments like doc string and bullets. It feels that it can be part of the same unit test and it would not require significant change of code.

Regards,
Milos

deyanp:
Sure, will do

deyanp:
and now with the rest as well ...
if ok, when can we have a new nuget version, so that we can use it in our slns?

TickSpec/FeatureParser.fs
```diff
@@ -91,12 +91,20 @@ let parseFeature (lines:string[]) =
         let bullets, table, doc =
             match step.Item with
             | Some (BulletsItem b) ->
-                b |> List.toArray |> Some, None, None
+                b
+                |> List.map (replace combination)
```
bartelink, line 95: Seq.map >> Seq.toArray

TickSpec/FeatureParser.fs
```diff
@@ -91,12 +91,20 @@ let parseFeature (lines:string[]) =
...
+                |> List.toArray |> Some,
+                None, None
             | Some (TableItem t) ->
                 let header = t.Header |> List.toArray
-                let rows = t.Rows |> Seq.map List.toArray |> Seq.toArray
-                None, new Table(header, rows) |> Some, None
-            | Some (DocStringItem d) -> None, None, Some d
+                let rows = 
+                    t.Rows 
+                    |> Seq.map (List.map (replace combination))
```
bartelink, line 102: replace intermediate usage of List with Seq or similar

deyanp: but it was already a List, I am not converting it, just mapping it ...

TickSpec/FeatureParser.fs
deyanp: Not sure I understand this one, b is string list, I am just List.mapping before the List.toArray which was already there ..

TickSpec/FeatureParser.fs
deyanp: I can do the following if that is what you mean:

`|> Seq.map (List.map (replace combination) >> List.toArray)`

deyanp: I doubt that will make any noticeable diff though performance-wise

TickSpec/FeatureParser.fs
bartelink: It doesnt need to be a cons list with lots of allocations - just because it started as a List does not mean it has to stay one

TickSpec/FeatureParser.fs
bartelink: I get that it makes no real difference perf-wise but List.map and Map.map etc are something that for me one only employs when the final materialised rendition of the data really needs to be a List.
So for this I'd use `|> Seq.map (Seq.map (replace combination) >> Seq.toArray)` on style grounds more than perf.
Kit Eason's pluralsight course on F# Functional Data Structures is pretty good to explain the deeper reasoning.
(Without looking at the rest of the code I'm wondering... Does the ultimate result actually need to be a seq of arrays - maybe a collect might make sense?)

deyanp: The ultimate result must be array of string arrays string [] []

Look, I have pushed something, now using Seq.map instead of List.map ... I would have never imagined that I can use Seq.map on a list, was using till now List.map on a list and Seq.map on a seq, and Array.map on array ... but if you thing this is better, then so be it, I am interested in having the placeholder replacement only ...

TickSpec/FeatureParser.fs
deyanp: I have pushed sth - is this is what you expect?

bartelink: Yes, looks good!

TickSpec/FeatureParser.fs
bartelink: Yes, looks good now.

Aside: probably not, but query comprehension syntax can sometimes work well for this sort of transformation i.e.
```
let rows =
    let replaceInRow xs = [| for x in xs -> x |> replace combination |]
    [| for r in t.Rows -> replaceInRow r |]
```
In this instance, that's probably breaking the flow of the code compared to how the surrounding code is written though

deyanp: Thanks for sharing that option too.

Just a final question - why do you insist on using Seq.map on a list instead of List.map - due to performance/memory?

bartelink: TL;DR yes but I'll give some more detail...

I generally default to not materialising intermediates of mappings - leaving them as lazy (seq) for as long as possible. In C# Linq, that manifests as a suspicion of `ToList` and things like that.

In some cases, you want to avoid double enumeration though, which brings a need/desire to compute and freeze it - i.e. toArray / toList etc. However `Seq.cache` can often be the half-way-house answer.

While F# List and Map, being thread safe persistent data structures are very good default types to use for lots of reasons, I am often conscious that Array and HashSet are much more compact in terms of memory usage - both in terms of amount of allocations, but also memory locality wrt NUMA/caching behaviors. You don't want to let that dictate your design - the legibility of the code trumps most things until you actually have a Benchmark.net proving there is a real problem (and even in that case, you want it to be significant in terms of the whole system)

As mentioned, I highly recommend Kit Eason's F# Functional Data Structures PluralSight course - its 4h but its fantastic and gives one a feel for all these sorts of things - well worth the time.

Whether one uses `Seq.map f >> Seq.map g`  over `Seq.map (f >> g)` for me is a more subjective thing - whatever makes the code cleanest to read generally dictates it for me - the important thing is that I'm not materialising things in either case

(In Equinox and its code examples, you'll see me defaulting to List and Map quite a bit as stuff needs to be threadsafe, but quite often I'll use an Array in preference to either of those for memory consumption reasons for the `State` type - this is driven by the fact that it occupies memory in a `MemoryCache`, in which case you want it to be dense (and as that makes things long-lived from a GC perspective, having less memory blocks involved is also a win for overall perf.))

mchaloupka:
@bartelink Do you have any further comment? If not, I will merge it and create a minor release so it is ready to be used.

bartelink:
No, in general the impl looks good and has tests so no objections from me (I've not run it or anything though ;))

PR #51, opened 2021-06-14
Enable nuget publishing of Xunit wiring

mchaloupka:
I have done changes necessary to publish Xunit wiring publishing.

The change is that I am calling `dotnet pack` on the whole solution. Therefore, I had to mark all the projects as non-packable and only the selected ones are marked as packable. I have extracted shared properties for the to-be published projects to a new props file.

It seems that the `pack` is now triggering build. Not sure why that happens. It does not matter much in our case so I just ensured that the tests will be executed without building after packing, so it should be running using the packed libraries.


Wiring/TickSpec.Xunit/TickSpec.Xunit.fsproj
```diff
@@ -1,9 +1,19 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../../netfx.props" />
+  <Import Project="../../nuget.props" />
```
bartelink, line 4: cant these be done as Directory.build.props ?

nuget.props
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation.  All Rights Reserved.  See License.txt in the project root for license information. -->
+<Project>
+  <PropertyGroup>
+    <IsPackable>true</IsPackable>
+    <Authors>Phillip Trelford, Ruben Bartelink, Milos Chaloupka</Authors>
+    <Copyright>Copyright © 2010-21 Phillip Trelford, Ruben Bartelink, Milos Chaloupka</Copyright>
+    <PackageProjectUrl>https://github.com/fsprojects/TickSpec</PackageProjectUrl>
+    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
+    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
```
bartelink, line 10: also add packagelincense like this https://github.com/jet/propulsion/blob/master/Directory.Build.props#L8

bartelink (commented):
How about putting the two packaged projects into a `/src` folder (and maybe putting others under `/examples`) and then putting an `<IsPackable>true` in that folder's Directory.Build.props ?

You could also inline the bulk of the nuget.props into there

I dont think doing a global pack makes sense, esp when we only have 2 nugets to generate. [Here's a way I did it](https://github.com/jet/propulsion/blob/master/build.proj#L15-L24). While it'd be ideal to test with identical binaries, in general I think letting `dotnet pack` and `dotnet test` both trigger incremental builds is not the end of the world.

For the sourcelink, also should add https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/#deterministic-builds ?

nuget.props
mchaloupka: Are you sure that the way you are proposing works? According to [this](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets), the following two values are supported `PackageLicenseFile` and `PackageLicenseExpression`. We can change it from file to expression though.

Wiring/TickSpec.Xunit/TickSpec.Xunit.fsproj
mchaloupka: Yes. However, I wonder whether it is not better to be explicit and mark the projects we want to pack.

TickSpec/TickSpec.fsproj
```diff
@@ -1,28 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../netfx.props" />
+  <Import Project="../nuget.props" />
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
-    <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
+    <RootNamespace>TickSpec.Xunit</RootNamespace>
```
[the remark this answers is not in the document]
mchaloupka: Yeah, I will get rid of it. (It is also incorrect)

mchaloupka:
> How about putting the two packaged projects into a `/src` folder (and maybe putting others under `/examples`) and then putting an `<IsPackable>true` in that folder's Directory.Build.props ?
> 
> You could also inline the bulk of the nuget.props into there
> 

We have also unit tests. I kind of like the explicit approach. On the other hand, the real problem is only with some of the examples project. The unit test projects are automatically marked as not packable. I will experiment with it to remove the need of the extra `nuget.props` file.

> I dont think doing a global pack makes sense, esp when we only have 2 nugets to generate. [Here's a way I did it](https://github.com/jet/propulsion/blob/master/build.proj#L15-L24).

Can you elaborate what is the advantage of your approach? To me it feels that it should be preferred to pack in one go, especially as there are references between the projects.

> While it'd be ideal to test with identical binaries, in general I think letting `dotnet pack` and `dotnet test` both trigger incremental builds is not the end of the world.

Yeah, I believe so.
 
> For the sourcelink, also should add https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/#deterministic-builds ?

Yeah, why not.

bartelink:
> Can you elaborate what is the advantage of your approach? To me it feels that it should be preferred to pack in one go, especially as there are references between the projects.

Not really ;) The main thing for me is to be able to trigger a given action from the command line and/or from a CI driver consistently.

If I dont have a monolithic fake script, then using the standard SDK-provided `dotnet pack` etc is that thing.

This allows one to e.g. run the tests on 3 environments, but explicitly trigger the pack only on one.

Of course e.g. the way I have the tests there mean that you dont parallelize the test runs or builds as one can do when you use a .sln (in Equinox I did that: https://github.com/jet/equinox/blob/master/build.proj#L32)

But ultimately it's all subjective - you can achieve the same effects no matter how you do it.

It might be possible to do a `dotnet build` on the sln and then do a `dotnet pack` and/or some `dotnet test`s with `--no-build` or whatever it is to force it not to rebuild even if ordinarilyit might (e.g. due to settings discrepancies)

Wiring/TickSpec.Xunit/TickSpec.Xunit.fsproj
bartelink: I agree Explict is better than Implicit in general. However, I think D.B.P is a well established convention at this point, as are all `dotnet *` commands.

nuget.props
bartelink: the expression flows to the packed nupkg and helps with some things.
there is awarning that dotnet pack gives if you dont do it I think

mchaloupka:
> It might be possible to do a dotnet build on the sln and then do a dotnet pack and/or some dotnet tests with --no-build or whatever it is to force it not to rebuild even if ordinarilyit might (e.g. due to settings discrepancies)

Yes, and we are not so far from it. It was not originally possible to run `dotnet test` on solution and that is the reason, why it is so specific now. I believe that with a little of work, it should be possible to run it on solution (just the xunit projects needs to have updated frameworks on non-windows platforms I believe).

mchaloupka:
For simplicity, I disabled running .net 4 tests under linux. We want to get rid of the framework eventually so it is probably not worth the effort to configure the xunit projects to not target .net 4 on linux.

Anyway, it should be complete. What do you think?

bartelink:
Yes :shipit:

PR #54, opened 2022-06-04
Migration to Github Actions

sergey-tihon:


sergey-tihon:
@mchaloupka let's move our discussion from gutter to this PR.

Please review, this PR should be already in good shape
- Actions instead of my Appveyor
- main for build and test (master and PR to master)
- new macOS target for testing (artifacts still from windows builds)
- release flow (with manual trigger) for NuGet package publishing (not tested)
- NUGET_KEY secret migrated from AppVeyor
- Test and example project migrated .NET 5 to .NET 6
- FAKE updated to latest version and FAKE dependencies also constrained by .NET 6 runtime
- new build status badge and cleanup of build dependencies

If evethiing OK we can merge and remove Appveyor webhooks from repo.

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    - master
+
+jobs:
+
+  build:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macOS-latest]
```
mchaloupka, line 18: Considering that `macOS` build consumes 10x times more free minutes than linux build, should we keep only windows and ubuntu one?

If we get rid of .NET 4.x at some point, I would remove even windows build as it will stop bringing any value. What do you think?

TickSpec.sln
```diff
@@ -17,50 +17,46 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "CommandLine", "CommandLine"
...
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Interactive", "Examples\ByStyle\Interactive\Interactive.fsproj", "{BA6F3A69-34F3-4D85-8804-DD354F8E17FA}"
 EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Attributes", "Examples\ByStyle\Attributes\Attributes.fsproj", "{B3AA7D1B-E83A-4C9B-97CF-12293AF46E10}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Attributes", "Examples\ByStyle\Attributes\Attributes.fsproj", "{B3AA7D1B-E83A-4C9B-97CF-12293AF46E10}"
 EndProject
-Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Functional", "Examples\ByStyle\Functional\Functional.fsproj", "{1D2CA398-6F02-41B1-BAA5-E63DC15508A4}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Functional", "Examples\ByStyle\Functional\Functional.fsproj", "{1D2CA398-6F02-41B1-BAA5-E63DC15508A4}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "._", "._", "{51B6BE98-A176-4315-83F5-54800BC03EB4}"
 	ProjectSection(SolutionItems) = preProject
```
mchaloupka, line 35: I wonder, should we fix this folder or get rid of it? I do not have strong opinion except of not wanting to have the folder containing only some of the items.

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Build & Tests
+      run: dotnet fake run build.fsx -t "Test"
+      env:
+        BUILD_NUMBER: ${{github.run_number}}
+    - name: Upload NuGet Packages
+      if: matrix.os == 'windows-latest'
+      uses: actions/upload-artifact@v3
+      with:
+        name: nuget
+        path: packed_nugets/*.nupkg
```
mchaloupka, line 39: Include also symbols please `*.snupkg`. Or let the build handle it and upload everything in the `packed_nugets` folder.

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Upload Release Binaries
```
mchaloupka, line 40: Do we need to upload these? It may be weird to upload only netstandard 2 binaries (probably caused by how we build things). We should either fix it to include also .NET 4.5.2 binaries or not upload it at all.

mchaloupka (commented):
Thanks for doing this, only a few minor comments. Otherwise I believe that it is good to go and it would be nice to have the actions configured on the repository level and get rid of account-level appveyor to manage it.

TickSpec.sln
sergey-tihon: It depends on IDE you use, if project maintainers do not need folder - we can delete.

.github/workflows/main.yml
sergey-tihon: At the moment we do not have issues with `free minutes` in `fsprojects` org.  I think that we can keep all 3 for now to increate test coverage and ensure that it is easy to start contributing from any platform. 

If we face issue with `free minutes` on org level, we will deal with it in the future (for example remove macOS from repos that consume largest amount of minutes)

.github/workflows/main.yml
sergey-tihon: I hardly imaging how binaries may be used in usual workflow, so lets remove

.github/workflows/main.yml
sergey-tihon: Fixed

.github/workflows/main.yml
mchaloupka: Agree.

TickSpec.sln
mchaloupka: Please remove this folder from sln and then we can merge it.

sergey-tihon: Done, I also `disabled` AppVeyor web hook on the repo

PR #57, opened 2022-07-22
Fix test execution

mchaloupka:
Fixes test execution from IDE. Enables test execution from a deserialized scenario.


bartelink:
@mchaloupka FAKE and I [don't get along](https://github.com/fsprojects/FSharp.AWS.DynamoDB/pull/55) - Any scripting I'd do would be msbuild centric and/or in line with gtihub.com/jet (although that's on ADO, not GHA). In other words, best of luck with the quest ;P

Issue #55, opened 2022-06-04
"Missing step definition" when applying multiple step attributes to single step


jcoliz:
Hi. In the comments for issue #25, @michalkovy notes that it is possible to allow a single step to function both as (say) a Given and When step.

> We could potentially have such attribute but you can do the same today by writing:

```f#
let [<Given; When; Then>] ``I have defined a step definition`` () = ()1
```

However, this is not working for me today in 2.0.2.

This is what I have:
```f#
let [<Given; When>] ``user launches site`` (page: IPage) (uri: Uri) = 
    page.GotoAsync(uri.ToString()) |> Async.AwaitTask |> Async.RunSynchronously
```
```
Scenario: Root loads OK
    When user launches site
    Then page loaded ok

Scenario: Home page loads OK via navbar
    Given user launches site
    When clicking Home on navbar
    Then page title reads "Home"
```
This is what TickSpec has to say:
```
    TickSpec.StepException : Missing step definition on line 4 in Site is alive and healthy
	"When user launches site"
```
Note that the Given step still works fine. 

I can work around it by duplicating the step, but yuck.

This workaround does work:
```f#
let [<When>] ``DUPLICATE user launches site`` (page: IPage) (uri: Uri) = 
    ``user launches site`` page uri
```
```
Scenario: Root loads OK
    When DUPLICATE user launches site
    Then page loaded ok
```

jcoliz:
@mchaloupka Thanks!

PR #59, opened 2022-07-22
Allow multiple step types on a single method

mchaloupka:
This resolves https://github.com/fsprojects/TickSpec/issues/55

### DIFF
--- a/TickSpec/TickSpec.fs
+++ b/TickSpec/TickSpec.fs
@@ -134,7 +134,6 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
                         | step -> step
                     p,a,sm
                 )
-                |> Seq.distinctBy (fun (p,a,m) -> p)
             )
             |> Seq.fold (fun (gs,ws,ts) (p,a,m) ->
                 match a with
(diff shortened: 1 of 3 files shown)


TickSpec/TickSpec.fs
```diff
@@ -134,7 +134,6 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
                         | step -> step
                     p,a,sm
                 )
-                |> Seq.distinctBy (fun (p,a,m) -> p)
```
bartelink, line 137: On the face of it, this seems a reasonable change.

It seems strange though - the sort of thing that could have been added to shut up some condition during development for some subtle case.

But, dropping random matching steps doesnt make any sense to me

Surprised more than one tag of each kind (G/W/T) to be permissible though - that should probably be restricted via AttributeUsage?

Not really sure what test cases I'd even put around it as valid ones...

Don't have an objection to the PR as it stands though personally

mchaloupka: > It seems strange though - the sort of thing that could have been added to shut up some condition during development for some subtle case.

Agree. It seems that it is there from beginning of time. I was trying to think what it will break, but it just does not make sense to me and it will potentially just hide issues. Like the one we have in tests. We should rather fail tests when there is ambiguity than to hide it and have "random" behaviour depending on the order how the attributes are returned by reflection.

> Surprised more than one tag of each kind (G/W/T) to be permissible though - that should probably be restricted via AttributeUsage?

I think that it potentially makes sense. You can support different wording of the step and map it to the same method (considering that the order of argument is the same).

> Don't have an objection to the PR as it stands though personally
👍

bartelink: >I think that it potentially makes sense. You can support different wording of the step and map it to the same method (considering that the order of argument is the same).

Don't know; while I can imagine it making for a less convoluted regex sometimes, it definitely smells like a misfeature (but I can't say I've looked at how other impls approach such scenarios).

PR #67, opened 2024-01-23
Implement injection for events (fixes #56)

### LINKED ISSUE #56: Events (BeforeStep, AfterStep, AfterScenario) can't access arguments
Currently it's not possible to use shared arguments in Events like `BeforeStep`, `AfterStep`, `AfterScenario` (I don't think it makes sense for `BeforeScenario`)

For example this results in an error:

```
[<BeforeScenario>]
let setup () =
    { Server = TestServer.Start() }

[<AfterScenario>]
let teardown ctx =
    ctx.Server.Stop()
```

```
System.Reflection.TargetParameterCountException
Parameter count mismatch.
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at TickSpec.ScenarioRun.invoke(IInstanceProvider provider, MethodInfo m, Object[] ps)
   at TickSpec.ScenarioRun.invokeEvents@171-1.Invoke(MethodInfo mi)
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source) in D:\a\_work\1\s\src\fsharp\FSharp.Core\seq.fs:line 497
   at TickSpec.ScenarioRun.generate[a,b](IEnumerable`1 events_0, IEnumerable`1 events_1, IEnumerable`1 events_2, IEnumerable`1 events_3, IDictionary`2 parsers, a scenario, IEnumerable`1 lines, FSharpRef`1 serviceProviderFactory, Unit unitVar0)
   at <StartupCode$TickSpec>.$TickSpec.action@185.Invoke(Unit arg40@)
   at <StartupCode$TickSpec>.$TickSpec.GenerateScenarios@187-1.Invoke()
```

I know I can just use a mutable value instead, but then what's the point of using arguments at all? I think it would be great to be able to have the same arguments available in the events as in the step definitions.

mchaloupka:


mchaloupka:
This will resolve https://github.com/fsprojects/TickSpec/issues/56.

Examples/ByFeature/FunctionalInjection/Logger.fs
```diff
@@ -0,0 +1,23 @@
...
+let [<BeforeStep>] beforeStep (previousMessages: LoggerContext) =
+    { Messages = "Before step" :: previousMessages.Messages  }
+
+let [<AfterStep>] afterStep (previousMessages: LoggerContext) =
+    { Messages = "After step" :: previousMessages.Messages  }
+
+let [<AfterScenario>] afterScenario (previousMessages: LoggerContext) =
+    let allMessages =
+        "After scenario" :: previousMessages.Messages
+        |> List.rev
```
bartelink, line 22: There's no value in having the interstitial reversed state manifest as a linked list of nodes, so:
allMessages |> Seq.rev |> Seq.iter Console.Writeline
(I'd prob write `for m in Seq.rev allMessages do Console.WriteLine m` myself)

mchaloupka: Good point.

TickSpec/ScenarioRun.fs
```diff
@@ -18,10 +18,9 @@ let getInstance (provider:IInstanceProvider) (m:MethodInfo) =
 
 /// Invokes specified method with specified parameters
 let invoke (provider:IInstanceProvider) (m: MethodInfo) (ps: obj array) =
-    let injectionArgs = 
-        let pars = m.GetParameters()
-        let a = ps.Length
-        Array.sub pars a (pars.Length - a)
+    let injectionArgs =
```
bartelink, line 21:
```
let args = [|
    yield! ps
    for p in m.GetParameters() |> Seq.skip ps.Length do
        provider.GetService p.ParameterType |]
```

mchaloupka: Looks good, will use.

Issue #45, opened 2021-02-07
Big scenarios execution time


AndrewRybka:
Hi,

It takes ~111 ms to run all the 3 scenarios from the ByFramework/xUnit example when I run it as a standalone project:

![image](https://user-images.githubusercontent.com/1169619/107161983-f9b8eb80-69a8-11eb-8f59-aa91ee09382e.png)

But when I move that example to one of my test projects with 600+ other tests, the execution time raises to 5+ seconds, and each scenario is executed for more than a second:

![image](https://user-images.githubusercontent.com/1169619/107162550-af396e00-69ac-11eb-9af3-66fe7fa51c31.png)

Is there a way to reduce scenario execution time in the second case?

Thanks

mchaloupka:
Hi,

are you able to share a project where it happens so we can reproduce it?

Potentially, are you able to run a profiler to see in which method it is spending most time?

Thanks,
Milos

mchaloupka:
Hi @AndrewRybka,

are you able to run a profiler to see what is the issue?

Thanks,
Milos

AndrewRybka:
Hi Milos,

Sorry for the delay with the answer.

Unfortunately, I can't share the initial project.

But I generated a project with ~2K unit test with a single assert in each (XUnit facts, theories, and FsCheck properties) in an attempt to replicate the issue. With that project it takes 900-1000 ms to execute scenarios from the ByFramework/xUnit example. With each scenario executed for 200-400 ms. And when I add additional scenarios, the overall execution time grows proportionally. So it seems that the number of tests in the project affects the execution time of each scenario. Here is the project with 2K tests:
[ByFramework.XUnit.zip](https://github.com/fsprojects/TickSpec/files/7782019/ByFramework.XUnit.zip)

Also, I profiled the initial project while running only tests from the ByFramework/xUnit example. Here is the results overview:
[snapshot.txt](https://github.com/fsprojects/TickSpec/files/7781850/snapshot.txt)

I can send the whole profiling snapshot, created by dotTrace, via email.


Thanks

plainionist:
i observe similar behavior: the more tests (bigger scenarios) i add the slow the execution gets. any insights meanwhile what the root cause could be and how to improve?

mchaloupka:
I have created a pull request with improvements: https://github.com/fsprojects/TickSpec/pull/69. That should improve it significantly.

plainionist:
The improvement in execution time is HUGE in my case - thx a lot!

[...]

PR #69, opened 2024-01-24
Issue #45: Improve execution time in large assemblies

mchaloupka:
The usage of `seq` caused the reflection to be done repetitively for all operations. I have improved the execution time of the tests in the provided project (see issue) from 12 seconds to 1.4 seconds just by making this tiny change of materializing the sequences into lists. I had to materialize it in two places:
* The moment where we filter out all methods that are attributed by one of the event attributes.
* The place where we select events in-scope for a given scenario

### DIFF
--- a/TickSpec/ScenarioGen.fs
+++ b/TickSpec/ScenarioGen.fs
@@ -481,10 +481,7 @@ let defineStepMethod
 let defineRunMethod
     (scenarioBuilder:TypeBuilder)
     (providerField:FieldBuilder)
-    (beforeScenarioEvents:MethodInfo list,
-     afterScenarioEvents:MethodInfo list,
-     beforeStepEvents:MethodInfo list,
-     afterStepEvents:MethodInfo list)
+    (beforeScenarioEvents,afterScenarioEvents,beforeStepEvents,afterStepEvents)
     (stepMethods:MethodBuilder seq) =
     /// Run method to execute all scenario steps
     let runMethod =
@@ -496,8 +493,8 @@ let defineRunMethod
     let gen = runMethod.GetILGenerator()
 
     // Emit event methods
-    let emitEvents (ms:MethodInfo seq) =
-        ms |> Seq.iter (fun mi ->
+    let emitEvents =
+        Seq.iter (fun (mi:MethodInfo) ->
             if mi.IsStatic then
                 gen.EmitCall(OpCodes.Call, mi, null)
             else
--- a/TickSpec/ScenarioGen.fs
+++ b/TickSpec/ScenarioGen.fs
@@ -481,11 +481,11 @@ let defineStepMethod
 let defineRunMethod
     (scenarioBuilder:TypeBuilder)
     (providerField:FieldBuilder)
-    (beforeScenarioEvents:MethodInfo seq,
-     afterScenarioEvents:MethodInfo seq,
-     beforeStepEvents:MethodInfo seq,
-     afterStepEvents:MethodInfo seq)
-    (stepMethods:seq<MethodBuilder>) =
+    (beforeScenarioEvents:MethodInfo list,
+     afterScenarioEvents:MethodInfo list,
+     beforeStepEvents:MethodInfo list,
+     afterStepEvents:MethodInfo list)
+    (stepMethods:MethodBuilder seq) =
     /// Run method to execute all scenario steps
     let runMethod =
         scenarioBuilder.DefineMethod("Run",
--- a/TickSpec/TickSpec.fs
+++ b/TickSpec/TickSpec.fs
@@ -88,6 +88,7 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
             xs
             |> Seq.filter (fun m -> m |> isMethodInScope feature scenario)
             |> Seq.map (fun (_,_,_,e) -> e)
+            |> Seq.toList
         events
         |> fun (ea,eb,ec,ed) -> choose ea, choose eb, choose ec, choose ed
     new () =
@@ -144,7 +145,10 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
             ) ([],[],[])
 
         let filter (t:Type) (elements:(string list * string list * string list * MethodInfo) seq) =
-            elements |> Seq.filter (fun (_,_,_,m) -> null <> Attribute.GetCustomAttribute(m,t))
+            elements
+            |> Seq.filter (fun (_,_,_,m) -> null <> Attribute.GetCustomAttribute(m,t))
+            |> Seq.toList
+
         /// Step events
         let events = methods |> filter typeof<EventAttribute>
         let beforeScenario = events |> filter typeof<BeforeScenarioAttribute>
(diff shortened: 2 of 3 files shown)


TickSpec/ScenarioGen.fs
```diff
@@ -481,11 +481,11 @@ let defineStepMethod
 let defineRunMethod
     (scenarioBuilder:TypeBuilder)
     (providerField:FieldBuilder)
-    (beforeScenarioEvents:MethodInfo seq,
-     afterScenarioEvents:MethodInfo seq,
-     beforeStepEvents:MethodInfo seq,
-     afterStepEvents:MethodInfo seq)
-    (stepMethods:seq<MethodBuilder>) =
+    (beforeScenarioEvents:MethodInfo list,
```
bartelink, line 484: I'd argue these should become arrays - more emory efficient and not a linked list so potentially better cache locality
Yes technically they are mutable
Eventually there will be an `iarray` alias and can flip to that

TickSpec/TickSpec.fs
```diff
@@ -88,6 +88,7 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
             xs
             |> Seq.filter (fun m -> m |> isMethodInScope feature scenario)
             |> Seq.map (fun (_,_,_,e) -> e)
+            |> Seq.toList
```
bartelink, line 91: to build this into an array, best to use an array comprehension
```
[| for _, _, _, e as x in xs ->
       if isMethodInScope feature scenario x then
           e |]
```

bartelink:
Wow, nice work to figure out such a productive change

TickSpec/ScenarioGen.fs
mchaloupka: I have removed the type from there to make it agnostic as long as it is anything with an iterator.

TickSpec/TickSpec.fs
mchaloupka: Looks good.

mchaloupka:
I have done further improvements to scan through the methods less times (it cuts the time in half again).. However, as some of the tests are failing, so will look into fixing that tomorrow.

TickSpec/TickSpec.fs
```diff
@@ -98,69 +111,103 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
...
-                    let tags', features', scenarios' = getScope attributes
-                    tags@tags' |> List.filter (not << String.IsNullOrEmpty),
-                        features@features' |> List.filter (not << String.IsNullOrEmpty),
-                            scenarios@scenarios' |> List.filter (not << String.IsNullOrEmpty),
-                                m
+            |> Seq.collect (fun t -> t.GetMethods())
+        StepDefinitions(methods)
+    internal new (methods:MethodInfo seq) =
+        let categorizedMethods =
+            let getScope attributes =
```
bartelink, line 120: much more efficient to do it this way
probably also the add calls should have the filtering at this point too
```
          let getScope attributes =
                let t, f, s = ResizeArray(), ResizeArray(), ResizeArray()
                let add (x: ResizeArray) s = if not (String.isNullOrEmpty s) then x.Add s
                for x in attributes }> Seq.cast<StepScopeAttribute> do
                    add t x.Tag
                    add f x.Feature
                    add s x.Scenario
                t.ToArray(), f.ToArray(). s.ToArray()
```                
if the reversed order is necessary, `x.Add` can become `x.Insert(0,x)`

TickSpec/TickSpec.fs
```diff
@@ -98,69 +111,103 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
...
+                typeof<BeforeScenarioAttribute>; typeof<AfterScenarioAttribute>
+                typeof<BeforeStepAttribute>; typeof<AfterStepAttribute>
+                typeof<ParserAttribute>
+            |]
+
+            // Initialize the attribute map
+            attributes |> Array.iter (fun attrType -> attributeMap.[attrType] <- List.empty)
+
+            // Iterate through all methods
+            methods |> Seq.iter (fun mi ->
```
bartelink, line 142: for FTW
```
for mi in methods do
    for attr in mi.GetCustomAttributes true do
       match .... with
       | None -> ()
       | Some attrType ->
       
       <action>
```     
i.e. unindent for the normal case - this is considered idiomatic formatting and makes this far more legible without wide screen!

TickSpec/TickSpec.fs
```diff
@@ -98,69 +111,103 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
...
+                        if methodScope.ContainsKey mi |> not then
+                            let parentType = mi.DeclaringType
+                            if parentScope.ContainsKey parentType |> not then
+                                let parentScopeAttribute = parentType.GetCustomAttributes(typeof<StepScopeAttribute>,true)
+                                parentScope.[parentType] <- getScope parentScopeAttribute
+                            let tags', features', scenarios' = parentScope.[parentType]
+                            let methodScopeAttribute = mi.GetCustomAttributes(typeof<StepScopeAttribute>,true)
+                            let tags, features, scenarios = getScope methodScopeAttribute
+                            methodScope.[mi] <- (
+                                tags@tags' |> List.filter (not << String.IsNullOrEmpty),
```
bartelink, line 164: @ is very inefficient, as is rewriting with filtering, so the filtering moves into the gather
if these are coming back as arrays, Array.append is much better
(for ultimate efficiency you would pass in the accumulator into getScope but most of the win is already here and its clearer this way)

TickSpec/TickSpec.fs
```diff
@@ -98,69 +111,103 @@ type StepDefinitions (givens,whens,thens,events,valueParsers) =
...
-                ca
-                |> Array.map (fun a ->
-                    let p =
-                        match (a :?> StepAttribute).Step with
-                        | null -> m.Name
-                        | step -> step
-                    p,a,sm
-                )
+        let extractStepAttribute stepAttribute =
+            categorizedMethods.[stepAttribute]
```
bartelink, line 181:
```
            [| for _, _, _, m as method, attr in categorizedMethods[stepAttribute] ->
                let p =
                    match (attr :?> StepAttribute).Step with
                    | null -> m.Name
                    | step -> step
                p,method |]
```

TickSpec/TickSpec.fs
mchaloupka: I was thinking about changing it to arrays, but that would require bigger changes (or just changing them to list back). Considering how many values I expect here, the list should be absolutely fine and if not, then we should use set instead of array anyway. So, I will keep it to use the fold but will use the idea of accumulator and will move the filtering in the getScope method.

TickSpec/TickSpec.fs
mchaloupka: Considering that I don't care about efficiency so much here (there will be only a few values), I decided to leave the implementation similar with incorporated suggestions from the other comment. I also added the `add` method as you suggested, that is a great point. No reason to filter it afterwards.

TickSpec/TickSpec.fs
mchaloupka: Cannot use for-loop in constructor due to weird limitation of the older F# that is used in .NET Framework (see https://github.com/dotnet/fsharp/issues/13533). Had it using for loops originally :)

Have unindented the code. Good point, haven't been aware of this possibility.

TickSpec/TickSpec.fs
mchaloupka: Nice! Will use.

TickSpec/TickSpec.fs
bartelink: 👍 It all reads nice and clear now, and that's the main thing...

PR #81, opened 2026-03-22
Support escaped pipe (\|) in Gherkin table cells

deyanp:
## Summary

- Add support for `\|` escape sequences in Gherkin table cells, aligning TickSpec with the [Gherkin specification](https://cucumber.io/docs/gherkin/reference/#escaping) and other frameworks (Cucumber, SpecFlow)
- The `TableRowLine` parser now replaces `\|` with a placeholder before splitting on `|`, then restores literal `|` in the resulting cell values
- Existing behavior is fully preserved — tables without `\|` work exactly as before

## Motivation

Currently, it's impossible to include a literal pipe character (`|`) in a Gherkin table cell value, since `|` is always treated as a column delimiter. The Gherkin spec defines `\|` as the standard escape for this, and other BDD frameworks support it. This is useful when testing values that naturally contain pipes (e.g., KQL queries, shell commands, regex patterns).

## Test plan

- [x] Added `EscapedPipe.feature` with scenarios covering: no escapes (backward compat), single escaped pipe, leading/trailing pipes, multiple escaped pipes in one cell, and step tables with escaped pipes
- [x] Added `EscapedPipe_ParseLines` test verifying line-level parsing produces correct `TableRow` values with literal `|`
- [x] Added `EscapedPipe_ParseFeature` test verifying full feature parsing resolves escaped pipes in both Scenario Outline parameters and step table cells
- [ ] Run existing test suite to confirm no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/TickSpec.Tests/EscapedPipe.feature
+++ b/TickSpec.Tests/EscapedPipe.feature
@@ -1,0 +1,22 @@
+Feature: Escaped pipe in table cells
+
+    Scenario Outline: Values with escaped pipes are parsed correctly
+        Given a value <value>
+        Then the value is <expected>
+
+        Examples:
+            | value              | expected         |
+            | no pipe            | no pipe          |
+            | before\|after      | before\|after    |
+            | \|leading          | \|leading        |
+            | trailing\|         | trailing\|       |
+            | a\|b\|c            | a\|b\|c          |
+
+    Scenario: Step table with escaped pipes
+        Given a table with escaped pipes
+            | col1            | col2   |
+            | hello\|world    | normal |
+            | a\|b\|c         | x      |
+        Then the table cell 0,0 is hello|world
+        And the table cell 0,1 is normal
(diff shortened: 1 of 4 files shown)


mchaloupka:
Would you mind adding support for the full escaping logic as described?
https://cucumber.io/docs/gherkin/reference/#table-cell-escapin
> If you want to use a newline character in a table cell, you can write this as \n. If you need a | as part of the cell, you can escape it as \|. And finally, if you need a \, you can escape that with \\.

E.g. the code currently rewrites `\\|` into `\|` which is incorrect. It should be just `\` and then split.

mchaloupka:
Thanks for doing this work!

deyanp:
@mchaloupka when can we have this released?

mchaloupka:
Hi, sorry for the delay - just released it as 2.0.5 version

